PR #55, opened 2026-06-10
fix: validate lab_id against Docker tag grammar

### LINKED ISSUE #27: Build failed: invalid tag 'benchpress/DEV-v15:latest': invalid reference format
Hi, i have created two Labs Version 15 and Version 16 but when i click on button Actions- Build Image it returns
"invalid reference format "

For V16 i have added those apps

Image

Here is the result after build image 

Image

What can i do?

Venkateshvenki404224:
Fixes #27

## Problem

`build_lab_image` tags images as `benchpress/{lab_id}:latest`, but `lab_id` is free text. An ID like `TESTE V16` (space + uppercase) produces an invalid Docker reference and the build fails with the cryptic `invalid tag 'benchpress/TESTE V16:latest': invalid reference format`.

## Fix

- **`docker_manager.validate_lab_id()`** — canonical validator: lowercase alphanumerics with single `.`/`_`/`-` separators, max 64 chars (a strict subset of Docker's repository-name grammar).
- **`Lab.validate`** calls it, so bad IDs are rejected at save time with an actionable message — in both the desk form and the SPA (`lab.insert.error` is already rendered by `ErrorMessage` in `NewLab.vue`).
- **`build_lab_image()`** calls it too, covering labs created before this fix: the build now fails fast with `=== Build failed: Lab ID 'TESTE V16' is not valid: ... ===` in the build log instead of the docker error.
- **`build_lab`'s error path** now uses `frappe.db.set_value` instead of `doc.save()` — previously a validation error raised during the `except` handler's save would escape, leaving the lab stuck in its prior status and the build log without the failure line.
- UX hints added to the `lab_id` field (desk description + SPA help text).

No rename patch for existing bad labs: instance IDs are `md5(email + lab_id)` and container labels carry the old ID, so auto-renaming is riskier than letting owners rename manually (the build-log message tells them what to fix).

## Testing

- 2 new integration tests (`test_lab.py`): valid IDs insert, invalid IDs (`TESTE V16`, `DEV-v15`, `-leading`, `a..b`, 65 chars, …) raise `ValidationError`. Full suite: 56/56 pass.
- Browser-verified with agent-browser against the running stack:
  - Desk form: saving `TESTE V16` shows the validation dialog with the descriptive message.
  - SPA New Lab: submitting `TESTE V16` renders the same message in the form's error alert; `teste-v16` creates the lab and lands on /labs.
  - Legacy path: a pre-existing `TESTE V16` lab (inserted with validation bypassed) now ends Build Image with status `Error` and the clear message in the Build Log, no unhandled exception.
- `pre-commit run --all-files` passes.

Note (out of scope, observed while testing): on develop the SPA's `www/frontend.html` template expects a `boot` dict (incl. `csrf_token`) but `www/frontend.py` does not exist, so all SPA POSTs currently fail with CSRF `Invalid Request` — the boot-info adoption PR addresses this; SPA verification here was done with a CSRF token injected into the page.

### DIFF
--- a/benchpress/benchpress/doctype/lab/lab.json
+++ b/benchpress/benchpress/doctype/lab/lab.json
@@ -32,6 +32,7 @@
  ],
  "fields": [
   {
+   "description": "Lowercase letters, numbers and single '.', '_' or '-' separators, e.g. crm-lab",
    "fieldname": "lab_id",
    "fieldtype": "Data",
    "label": "Lab ID",
--- a/benchpress/benchpress/doctype/lab/lab.py
+++ b/benchpress/benchpress/doctype/lab/lab.py
@@ -4,6 +4,9 @@
 # import frappe
 from frappe.model.document import Document
 
+from benchpress.docker_manager import validate_lab_id
+
 
 class Lab(Document):
-	pass
+	def validate(self):
+		validate_lab_id(self.lab_id)
--- a/benchpress/benchpress/doctype/lab/test_lab.py
+++ b/benchpress/benchpress/doctype/lab/test_lab.py
@@ -1,7 +1,7 @@
 # Copyright (c) 2026, Venkatesh and Contributors
 # See license.txt
 
-# import frappe
+import frappe
 from frappe.tests import IntegrationTestCase
 
 # On IntegrationTestCase, the doctype test records and all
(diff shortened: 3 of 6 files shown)


Venkateshvenki404224:
Browser verification (driven with [agent-browser](https://github.com/vercel-labs/agent-browser) against the running dev stack):

**SPA — New Lab form**: submitting the exact ID from #27 (`TESTE V16`) is rejected with the validation message rendered in the form's error alert; the field now also shows a format hint. Re-submitting as `teste-v16` creates the lab and lands on /labs.

SPA validation error

**Desk — Lab form**: saving with `TESTE V16` shows the same message in the msgprint dialog.

Desk validation dialog

**Legacy lab path** (console-verified): a pre-existing `TESTE V16` lab now ends Build Image with status `Error` and this in the Build Log instead of the docker traceback:

```
=== Build started ===
=== Build failed: Lab ID 'TESTE V16' is not valid: use only lowercase letters, numbers and single '.', '_' or '-' separators (max 64 characters), e.g. 'crm-lab' or 'dev-v15'. ===
```

Screenshots live on the throwaway `screenshots/issue-27` branch; safe to delete after merge.

Issue #16, opened 2026-03-26
[Phase 1] Public browser access: HTTPS domain routing for spawned sites


Venkateshvenki404224:
## Summary

Currently, users access bench containers directly via WireGuard VPN IPs (e.g., `10.10.0.2:8000`). This works but requires every user to install and configure a WireGuard client.

Add Traefik as a reverse proxy so containers are accessible via subdomains like `bench1.yourdomain.com` with automatic HTTPS (Let's Encrypt).

## Proposed Behavior

- Each bench gets a subdomain: `<bench_name>.<base_domain>` (e.g., `crm-lab.benchpress.example.com`)
- Traefik handles TLS termination with auto-provisioned Let's Encrypt certificates
- No WireGuard required for web access — only for SSH
- Traefik routes traffic to the correct container based on the `Host` header

## Implementation Notes

- Run Traefik as a Docker container on the `benchpress` network
- Add Docker labels to bench containers during creation in `docker_manager.py`:
  ```python
  labels = {
      "traefik.enable": "true",
      "traefik.http.routers.<name>.rule": f"Host(`{bench_name}.{base_domain}`)",
      "traefik.http.routers.<name>.tls.certresolver": "letsencrypt",
      "traefik.http.services.<name>.loadbalancer.server.port": "8000",
  }
  ```
- Use the existing `base_domain` and `traefik_network` fields in BenchPress Settings (already defined but unused)
- Wildcard DNS record needed: `*.benchpress.example.com → server IP`

## Benefits

- Users can access Frappe sites in their browser without installing WireGuard
- HTTPS out of the box (no manual cert management)
- WireGuard remains useful for SSH access only
- Professional subdomain URLs for each bench

## References

- BenchPress Settings already has `base_domain` and `traefik_network` fields
- [Traefik Docker provider docs](https://doc.traefik.io/traefik/providers/docker/)

Venkateshvenki404224:
## Objective
Deliver the first BenchPress networking milestone: every spawned Frappe/ERPNext site should be accessible from a clean public HTTPS URL directly in the browser, without requiring VPN access.

## Product decision
BenchPress networking is split into two layers:

1. **Public browser access** — this issue/phase. Frappe Cloud-like URL access for demos, previews, training, and non-technical users.
2. **Private developer access** — later Headscale/Tailscale-compatible access for SSH, code-server, internal services, and team/device controls.

## User story
[...]

PR #62, opened 2026-06-13
feat: Phase 0 — bench reachable via Traefik over HTTP (#16)

Venkateshvenki404224:
## Phase 0 — Tracer bullet: one bench reachable via Traefik over HTTP

Thinnest end-to-end slice of issue #16. A deployed bench is reachable at
`http://<bench_name>.<base_domain>/`, routed by **Traefik** by `Host` header, with the
public URL shown on the dashboard. **HTTP only — no TLS, auth gate, toggle, or health** (those
are Phases 1–3). Touches every layer (infra → labels → deploy → site config → API → UI) to prove
the architecture before widening it.

### Changes
**Infra**
- `benchpress/config/traefik.yml` (new): static config — `web` entrypoint on `:80`, Docker
  provider on the `benchpress` network, `exposedByDefault: false`, insecure dashboard.
- `benchpress/config/docker-compose.yml`: add `benchpress-traefik` service (`traefik:v3.1`,
  `80:80` + `127.0.0.1:8090:8080`, docker.sock + traefik.yml mounted read-only).

**Backend**
- `benchpress/traefik_manager.py` (new): `ensure_traefik()` (idempotent) and
  `compute_bench_labels(bench, lab, settings)` — the single source of truth for a bench
  container's labels. The six `traefik.*` routing labels are emitted **only when
  `base_domain` is set**.
- `docker_manager.create_bench_container`: replace the hardcoded labels dict with
  `compute_bench_labels(...)`.
- `deploy_manager.deploy_bench`: after infra, `ensure_traefik()` when `base_domain` is set;
  set `bench.public_url`; set the site's `host_name` to the public URL so cookies/redirects use
  the canonical host. All gated on `base_domain` — a blank domain still deploys cleanly with no
  `traefik.*` labels.

**DocType + API + UI**
- `Bench Instance`: new read-only `public_url` field in a "Public Access" section;
  `api.get_benches` returns it.
- `LabDetail.vue`: clickable **Public URL** row in Connection Information (shown only when set).

**Tests**
- `tests/test_traefik_manager.py`: `compute_bench_labels` with and without `base_domain`
  (3 tests, passing).

### Verification (with `base_domain = lab.test`)
Traefik routing was proven end-to-end against a labeled container on the `benchpress` network
serving on port 8000, using the **real** `compute_bench_labels` output:

- `docker inspect` shows the expected `benchpress.*` + `traefik.*` labels.
- Traefik dashboard (`:8090`) lists the router `<name>@docker` with rule
  ``Host(`<name>.lab.test`)``, `web` entrypoint, service `<name>-svc` → `http://<ip>:8000`.
- `curl --resolve <name>.lab.test:80:127.0.0.1 http://<name>.lab.test/` → **HTTP 200**
  (`X-Forwarded-Server: benchpress-traefik`).
- SPA renders the clickable **Public URL** on the bench dashboard.

> Note: routing was verified with a lightweight stand-in container (no Frappe base image was
> cached locally for a full lab build); the SPA/API path was exercised against the running site
> with a seeded `Running` bench. The deploy-pipeline wiring is covered by the unit test + review.

**Traefik dashboard — HTTP router for the bench (Host rule → service port 8000):**

Traefik HTTP routers

**SPA — clickable Public URL in Connection Information:**

SPA Public URL

### Done when ✅
- [x] Bench reachable end-to-end over HTTP through Traefik by `Host` header
- [x] Right site responds; dashboard shows the URL
- [x] Deploy still succeeds with blank `base_domain` (labels omit all `traefik.*` keys)


Venkateshvenki404224:
## Phase 1 — HTTPS via Let's Encrypt (commit 231f805)

Layered on top of the Phase 0 work in this PR per `specs/phase-1-https-tls.md`. The bench public URL now serves over **HTTPS** with an auto-provisioned Let's Encrypt cert (TLS-ALPN-01), HTTP **301-redirects** to HTTPS, and a Settings toggle picks the LE **staging vs production** CA (staging default, to spare the prod rate limit).

**Changes**
- **Settings:** `acme_email` (Data) + `le_use_staging` (Check, default `1`).
- **traefik.yml:** `websecure` :443 entrypoint, HTTP→HTTPS redirect on `web`, `letsencrypt` ACME resolver with `tlsChallenge`.
- **docker-compose:** expose `443`, persist `acme.json` in named volume `benchpress-traefik-acme`.
- **compute_bench_labels:** route on `websecure` with `tls=true` + `certresolver=letsencrypt`.
- **deploy_bench:** `public_url` and site `host_name` become `https://`.

**One deliberate deviation from the spec** (flagged for review): the spec injects `email`/`caServer` into `traefik.yml` via `${ACME_EMAIL}`/`${ACME_CA_SERVER}` env vars. Traefik does **not** expand `${ENV}` in its *static* config file, and treats file vs CLI/env static config as mutually exclusive — I verified both empirically (extra CLI flags alongside `--configfile` were silently ignored). So instead `traefik_manager` **renders the ACME values into `traefik.yml` from Settings** at `ensure_traefik()` time — the same pattern `mariadb_manager` already uses for `mariadb.cnf`. The proxy is `--force-recreate`d only when the rendered config actually changes, so routine deploys don't bounce the shared proxy.

**Verification**
- 6/6 `test_traefik_manager` unit tests pass (label set + staging/prod CA rendering).
- Isolated Traefik run with the exact config: HTTPS `:443` → **200** over the default self-signed cert; HTTP `:80` → **301** `Location: https://…`; Traefik attempts ACME against the **staging** CA (confirming `le_use_staging` default). The only ACME failure was the expected "not a public TLD" for the fake test host — Traefik fell back to the self-signed cert and still served 200, exactly the spec's "local / no real cert" case.
- `pre-commit run --all-files` clean.

Venkateshvenki404224:
## Phase 1 verification screenshots + a follow-up fix

Adding the visual proof that was missing for the HTTPS phase (`docs/phase-1-verification/`), captured against the exact `traefik.yml` this PR ships.

### 1. Traefik routes the bench over HTTPS (websecure + TLS + LE resolver)
Dashboard router detail for a bench host: **WEBSECURE :443** entrypoint → router `Host(`crm-demo.lab.test`)` → **TLS: True**, **Certificate Resolver: letsencrypt**, status Success.

Traefik HTTPS router

### 2. TLS terminated at the proxy
Backend served over `https://…` through Traefik; it receives **`X-Forwarded-Proto: https`** and `X-Forwarded-Port: 18443`, i.e. Traefik terminated TLS and proxied the request.

HTTPS / TLS terminated

### 3. New BenchPress Settings fields
`ACME Email (Let's Encrypt)` and `Use Let's Encrypt Staging` (staging checked by default) in the Docker Configuration section:

Settings ACME fields

Full form for context:

BenchPress Settings form

Also covered by the curl checks from the spec: `http://…:80` → **301** `Location: https://…`, and `https://…:443` → **200** over Traefik's default self-signed cert (local / no real cert).

---

### Follow-up fix pushed (`6ea5c28`)
Capturing screenshot #3 surfaced a real bug: `le_use_staging` is a **new** Check field, and Frappe applies a Check default only at doc *creation* — so a `BenchPress Settings` singleton that predates the field reads **0**, and `_render_traefik_config` would pick the **production** CA, the opposite of the intended safe default. Added a `post_model_sync` patch (`benchpress.patches.set_le_use_staging_default`) that backfills `le_use_staging=1`. Verified post-migrate: the singleton reads `1` and the rendered config selects the **staging** CA.

PR #76, opened 2026-06-14
docs: backup-gated upgrade & rollback runbook (#47)

Venkateshvenki404224:
## What

First tracer-bullet slice of **[P1-07] #47 — Documented and scripted upgrade path for installed instances**.

Adds `docs/upgrading.md`: an operator runbook that upgrades a BenchPress install end-to-end (app code → schema → assets → restart → verify) with a **pre-upgrade backup gate** and a **tested rollback** procedure. Linked from the README guides list and the getting-started "What's Next?" table.

## Why a runbook first

This is the docs-first tracer for #47, mirroring how the #36 production-safety docs shipped (#70). It directly satisfies the issue's first acceptance criterion — *"one documented command/runbook upgrades an install end-to-end"* — and gives us the supported manual path immediately, while the automated `benchpress upgrade` command lands as a later slice.

## Acceptance criteria addressed

- [x] One documented command/runbook upgrades an install end-to-end.
- [x] Upgrade aborts safely if the pre-upgrade backup fails — **Step 0 is an explicit gate**: the upgrade must not proceed unless `bench backup` succeeds.
- [x] Policy defined for existing benches when a Lab's Frappe version changes — existing benches keep their image; rebuild is opt-in.
- [x] Rollback procedure documented — restores both code (recorded SHA) and data (Step 0 backup).

## Key decisions

- **Grounded in the real install model** from `getting-started.md` (host-bench `bench … migrate` / `bench build`), not the devops compose dev env.
- Scoped to the **control plane**; lab image rebuild is documented as an opt-in follow-up so upgrades stay non-disruptive to running benches.

## Still to come (flagged in the doc's "What's not automated yet")

- [ ] Scripted `benchpress upgrade` command chaining Steps 0–5 with abort-on-failure.
- [ ] Automatic rollback on mid-upgrade failure.
- [ ] Versioned `CHANGELOG` for multi-release upgrades.

## Testing

Docs-only change. Verified all internal doc links resolve and pre-commit hooks pass. (No code paths touched; browser feedback loop N/A.)

Closes #47 partially — keeping the issue open for the scripted slices above.

Refs #47

Issue #47, opened 2026-06-09
[P1-07] Documented and scripted upgrade path for installed instances


Venkateshvenki404224:
## Objective

Define how an existing BenchPress install is upgraded safely — app updates, schema migrations, lab image rebuilds, and eventually Frappe major-version bumps. This is the backbone of any paid "updates & support" offering: without it, every customer install drifts into its own unsupportable snowflake.

## Current Gap

There is no documented or scripted procedure to move a live install from version A to B: no pre-upgrade backup gate, no `bench migrate` orchestration, no guidance on rebuilding lab images against new Frappe branches, no rollback.

## Proposed Work

- Scripted upgrade flow: backup → app update → `bench migrate` → rebuild assets → restart → health verify.
- Define the policy for existing benches when a Lab's Frappe version changes (rebuild required? opt-in?).
- Rollback procedure (restore from the pre-upgrade backup — ties into #38).
- Versioned release notes / CHANGELOG so installs can be upgraded across multiple releases safely.

## Acceptance Criteria

- [ ] One documented command/runbook upgrades an install end-to-end.
- [ ] Upgrade aborts safely if the pre-upgrade backup fails.
- [ ] Rollback tested at least once on a real install.

## Priority
P1 — needed before the second customer install exists.

Venkateshvenki404224:
**RALPH iteration — first tracer bullet shipped: PR #76**

Shipped the smallest end-to-end slice of this issue: the **operator upgrade runbook** (`docs/upgrading.md`).

- Documents the full path — **Step 0 backup gate → app update → `bench migrate` → rebuild assets → restart → health verify** — plus a **rollback** that restores both code (recorded SHA) and data (the Step 0 backup).
- Step 0 is an explicit *gate*: the upgrade must not proceed unless `bench backup` succeeds, directly addressing *"upgrade aborts safely if the pre-upgrade backup fails."*
- Defines the **lab-image policy**: existing benches keep their image; rebuild is opt-in, so a control-plane upgrade stays non-disruptive.
- Grounded in the real install model (`getting-started.md`), linked from the README guides + getting-started "What's Next?".

**Decision:** docs-first tracer (mirrors the #36/#70 production-safety slice). It satisfies the *"one documented command/runbook upgrades an install end-to-end"* criterion now, while the automated command lands later.

Acceptance criteria progress:
- [x] One documented command/runbook upgrades an install end-to-end — **this PR**
[...]

PR #77, opened 2026-06-14
feat(upgrade): scripted backup-gated upgrade command (#47)

Venkateshvenki404224:
## What & why

Next RALPH slice for **#47 [P1-07] Documented and scripted upgrade path**. PR #76 shipped the manual runbook (`docs/upgrading.md`); this turns it into a scripted **`upgrade.sh`** that chains Steps 0–5 with the same **Step 0 backup gate** and **abort-on-failure**.

This directly advances the issue's acceptance criteria:
- [x] One documented **command**/runbook upgrades an install end-to-end — the *command* half now exists
- [x] Upgrade aborts safely if the pre-upgrade backup fails — Step 0 is a hard gate

## Key decisions

- **Shell script at the app repo root** (`apps/benchpress/upgrade.sh`), mirroring `setup.sh`. An upgrade is host-level orchestration (git + yarn + `bench restart`); a Frappe `bench` command would run *inside the process it restarts*, so a script — exactly parallel to `setup.sh` — is the consistent home.
- **Backup is a hard gate**: explicit error + non-zero exit if `bench backup` fails, so the upgrade never proceeds with nothing to roll back to.
- **`set -e` + an ERR trap** (installed only *after* the backup) abort on the first failure and print the recorded rollback SHA + backup pointer.
- **No auto-restore yet** — that stays a later slice, matching the runbook's "What's not automated yet".
- **`--dry-run`** prints every step (exercisable without a live bench); an optional `target-ref` does `git checkout <ref>` instead of `git pull --ff-only`.
- Doctor health check is best-effort (`|| warn`) since it may not exist on every install.

## Files

- `upgrade.sh` *(new, executable)* — backup-gated, abort-on-failure 5-step upgrade.
- `docs/upgrading.md` — new **Scripted upgrade** section + intro/footer cross-links; marked the scripted command done in "What's not automated yet".

## Verification

- `bash -n` clean.
- `--help`, missing-site, unknown-option, default-target dry-run, and target-ref dry-run all exercised against a fake bench layout — all five steps print; target-ref switches `pull` → `checkout`.
- `pre-commit` clean.
- ⚠️ Could not run a real upgrade in this environment (no live bench/site), consistent with how `setup.sh` is validated here.

## Notes for next iteration

- Automatic rollback on mid-upgrade failure.
- Versioned `CHANGELOG` for multi-release upgrades.

> Stacked on `docs/p1-47-upgrade-runbook` (PR #76) so the doc cross-links resolve; auto-retargets to `develop` once #76 merges.

Refs #47

### DIFF
--- a/docs/upgrading.md
+++ b/docs/upgrading.md
@@ -9,9 +9,10 @@ release to a newer one.
 > Frappe app and its site). Rebuilding the **lab images** your benches run on is
 > a separate, opt-in step covered in [Lab image policy](#lab-image-policy-after-an-upgrade).
 >
-> This is the documented runbook. A scripted `benchpress upgrade` command that
-> automates these steps (with the same backup gate and rollback) is the next
-> slice — see [What's not automated yet](#whats-not-automated-yet).
+> Prefer one command? `apps/benchpress/upgrade.sh` runs Steps 0–5 below with the
+> same backup gate and abort-on-failure behaviour — see
+> [Scripted upgrade](#scripted-upgrade). The manual steps remain the reference
+> for what each step does and for [rollback](#rollback).
 
 ---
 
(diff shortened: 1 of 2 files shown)


Venkateshvenki404224:
Consolidated: this intermediate tracer slice was #47 upgrade stack -> squashed into #80 (commit 359a072) on version-16. Closing — no changes lost.

Issue #94, opened 2026-07-05
[ISSUE-08] Remove MariaDB root password from exec command lines


Venkateshvenki404224:
## Source
Imported from Claude Web review shared by Venky on 2026-07-05.

## Priority
P1

## Branch
`fix/mariadb-pw-not-in-argv`

## Files
benchpress/mariadb_manager.py (execute_sql, backup_database_server)

## Scope
Replace mariadb -u root -p'{pw}' in the exec command with MYSQL_PWD passed via the exec environment= (or a chmod-600 defaults-extra-file). Same for the mariadb-dump backup command.

## Acceptance criteria
- [ ] No password appears in any command string passed to exec_run.
- [ ] ps inside the DB container during a query shows no root password.
- [ ] DB tests still pass.

## Verification
```bash
bench --site test_site run-tests --app benchpress --module benchpress.tests.test_mariadb_manager
```

## Risks
Medium — DB auth path; a wrong env var name breaks all SQL exec. Verify against a real MariaDB container, not just mocks.

## Do NOT touch
The base64 SQL-encoding scheme (keep it), the scoped-grant user creation logic.

## Agent/Ralph safety
Needs human review — agent drafts, human verifies auth against real MariaDB.

## Universal guardrails
- Never modify `cap_add` / `devices` / `privileged` / volume mounts in `create_bench_container`.
- Never touch `required_apps`, the MariaDB grant list, or add/remove `ignore_permissions=True`.
- Never call `remove_bench_volume`, `drop_site_database`, or `frappe.delete_doc(... force=True)` in a new code path without a human on the PR.
- Any change to `vpn_adapter.py` or the deploy/redeploy pipeline core is human-led; no autonomous merge.

Venkateshvenki404224:
Shipped via PR #109 (merged to version-16): MariaDB root password moved out of exec argv into MYSQL_PWD via exec environment — no longer visible in process lists.

PR #109, opened 2026-07-09
Implemented phase two

Venkateshvenki404224:
Phase one of issue #94 (`mariadb-pw-not-in-argv` spec): `execute_sql` — the seam all SQL routes through (`create_mariadb_user`, `drop_mariadb_user`, `drop_site_database`, `check_mariadb_health`) — now authenticates via `exec_run(environment={"MYSQL_PWD": root_pw})` instead of interpolating the root password into the argv string. Base64 temp-file scheme untouched.

**Verified live:**
- Smoke: `docker exec -e MYSQL_PWD=... benchpress_db mariadb -u root -e "SELECT 1"` returns 1 on mariadb:10.6.
- Full path: `SELECT SLEEP(8)` through `execute_sql` against managed `benchpress-mariadb`; concurrent `ps -ef` inside the container shows `mariadb -u root` with **no password in argv**; query exit 0.
- Regression test `test_execute_sql_passes_password_via_env_not_argv` asserts the sentinel password appears in no `exec_run` cmd and the SQL call carries `environment={"MYSQL_PWD": ...}`. Full `test_mariadb_manager` suite: 14/14 pass.

`backup_database_server` and `wait_for_mariadb` still use argv — phase 2 converts them.

**Needs human review before merge** (issue mandate: human verifies auth against real MariaDB).

### DIFF
--- a/benchpress/mariadb_manager.py
+++ b/benchpress/mariadb_manager.py
@@ -303,7 +303,8 @@ def wait_for_mariadb(
 	for _attempt in range(timeout // 2):
 		if direct:
 			exit_code, _output = container.exec_run(
-				cmd=["mariadb", "-u", "root", f"-p{root_pw}", "-e", "SELECT 1"],
+				cmd=["mariadb", "-u", "root", "-e", "SELECT 1"],
+				environment={"MYSQL_PWD": root_pw},
 			)
 			healthy = exit_code == 0
 		else:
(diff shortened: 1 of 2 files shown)


Venkateshvenki404224:
Phase two: backup_database_server and wait_for_mariadb direct poll now pass the root password via exec_run environment (MYSQL_PWD) instead of argv — zero password-in-argv sites remain (grep sweep for `-p'{root_pw}`/`-p{root_pw}` returns nothing).

Verify:
- `bench --site frontend run-tests --app benchpress --module benchpress.tests.test_mariadb_manager` — 15 tests OK, incl. new `test_backup_passes_password_via_env_not_argv`.
- Live backup ran against managed `benchpress-mariadb`: exit 0, `.sql.gz` written to `/var/lib/mysql/backups`; `ps -eo args` sampled during dump shows `mariadb-dump -u root --all-databases` — no password.
- Live direct poll `wait_for_mariadb(container=..., root_pw=...)` returned clean.

Per spec: PR needs human review before merge (auth verified live against MariaDB 10.6, but merge is human-gated).

PR #118, opened 2026-07-27
Implemented phase two

Venkateshvenki404224:
Rolling PR for the **deploy pipeline performance hardening** feature (`specs/in-progress/deploy-pipeline-performance/`, 5 phases). Phases one and two are in.

## Phase one — batch deploy and build log writes via `LogStream` (5decff7)

The log appender did a read-modify-write per line: O(n²) in bytes for n lines, since every append re-transferred the whole accumulated `longtext` in both directions. A Frappe image build emits 2,000–8,000 lines, so log persistence dominated build wall time.

`LogStream` replaces the appender closure — it buffers lines in memory and flushes them in one query-builder update that concatenates DB-side with no read-back. Flush triggers: 50 buffered lines, 2 seconds elapsed, or a terminal (`success`/`error`) line. The elapsed-time trigger is durability, not throughput: the bare `except` in `_deploy_bench` cannot catch a worker SIGTERM or the 1800s job timeout, so without it a killed build would lose its whole unflushed tail. The realtime publish moved after the persistence write, because the SPA reloads the persisted log when it sees a terminal line and would otherwise race the flush.

Measured on 3,000 lines: **61 writes instead of 9,000 round trips, 0.45s instead of 9.55s (21x)**.

## Phase two — bound log sizes and surface `bench build` failures (497d268)

See the phase-two comment below.

## Verification

- `bench --site frontend run-tests --app benchpress` → 222 tests green.
- `ruff check` / `ruff format --check` clean, `pre-commit run --all-files` 12/12.
- No DocType JSON changes in either phase, so no `bench migrate` needed. No frontend changes.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/benchpress/deploy_manager.py
+++ b/benchpress/deploy_manager.py
@@ -3,8 +3,11 @@
 
 import json
 import secrets
+import time
 
 import frappe
+from frappe.query_builder import DocType
+from frappe.query_builder.functions import Coalesce, Concat
 from frappe.utils.file_lock import LockTimeoutError
 from frappe.utils.synchronization import filelock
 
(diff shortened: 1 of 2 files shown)


Venkateshvenki404224:
**Phase two: bound logs and surface build failures** (497d268)

Two problems. Every string reaching a log or an exception was unbounded — one pathological output line (a base64 blob, a minified traceback) could be persisted whole, nothing capped the accumulated `longtext`, and command output was interpolated raw into exception messages, where `bench new-site` failure output can run to megabytes before it hits the `frappe.log_error` traceback. Separately, `bench build` discarded its `(exit_code, output)` return, so an asset-build failure passed silently and the deploy reported `Running`.

- `_tail(text, max_bytes)` plus two ceilings: `MAX_LOG_LINE_BYTES` (8 KB) and `MAX_LOG_BYTES` (512 KB). It tails rather than heads because the end of a command's output is where the failure is, and the elision marker is charged against the budget so the result never exceeds the cap.
- The per-line cap is applied inside `LogStream.__call__`, not at each call site — one choke point, so no caller can persist an unbounded line however the output was produced.
- `LogStream` tracks `written_bytes` and, once the next line would cross 512 KB, stops persisting and writes a single `=== log truncated ===` marker. The counter is exact without a read-back because the stream owns the document from creation; reading the stored size per line would undo phase one's batching.
- Realtime publishing is untouched. The cap bounds the stored audit trail only, so the SPA still receives every line the persisted log has dropped.
- `bench build`'s return is captured, its tailed output logged, and a non-zero exit raises — mirroring the `linkuser.sh` site directly above it. `_tail` also guards the two `raise Exception(f"... {output}")` sites.

## Verification

- `test_deploy_fails_when_bench_build_fails` was confirmed **red against pre-phase-two code** (`AssertionError: 'Running' != 'Error'`), green after. Plus `test_log_stream_caps_total_size` and `test_log_stream_tails_long_line`.
- `run-tests --module benchpress.tests.test_deploy_manager` 27 green; `--module benchpress.tests.test_api` 21 green (confirms the capped field still round-trips through `get_deploy_logs`); full app 222 green.
- Live through a real `LogStream` after `docker compose restart backend queue-long`: 8,002 lines (~1.7 MB) store **524,216 bytes** against the 524,288 cap with **exactly one** truncation marker, while all **8,002** realtime events still publish. A 100 KB single line stores at exactly **8,192** bytes with its tail intact and the head elided.
- SPA `/frontend/build-logs` renders an over-cap log with the marker shown once.

Not verified: a cold Frappe image build from the SPA. That path is unchanged by this phase and socket.io is not initialized in this environment, so the 8,002-line live stream above stands in for the publish-count contract.

PR #150, opened 2026-08-19
Implemented phase three

Venkateshvenki404224:
Deploy pipeline speed and reliability — phases 1, 2, 3 and phase 4's timeout half of the `deploy-pipeline-speed` spec, shipped out of numbered order for the demo, plus the fixes proving them end to end.

## What this branch does

- **Phase 2 — image cache redesign** (`0e3be8e`): static per-lab tags (`benchpress/<lab_id>:lab`), deploy fails fast on a missing image instead of silently building for 30-45 min, staleness reset when a Ready lab's spec changes.
- **Phase 1 — drop the per-bench volume** (`81331d0`): container creation no longer copies the whole 4-20 GB bench into a named volume. Measured live: container step 91s → 0.4s.
- **Phase 3 — assets never build in the container** (`c54fecc`): the dist-dir probe false-positived on SPA-style apps and re-ran `bench build` live on every deploy of such labs; assets ship in the image.
- **Phase 4 (timeouts)** (`c54fecc`, `2f1f8c9`): deploy/redeploy job ceiling raised to 2h to cover the real worst case; the site step drops a half-created database first so a killed run can't wedge an instance permanently.
- **Chown removal** (`df77ccc`): linkuser.sh's recursive chown was a structural no-op (the tenant user is the renamed uid-1000 frappe user) costing 8-30 min per deploy; the deploy now pushes the app's linkuser.sh into containers so the fix reaches already-built images.
- **Instance TLS + public URLs** (`81331d0`): instance Traefik routes name their certificate (the old `tls: {}` served Traefik's self-signed default → Cloudflare 526); sites are named `<instance-id>.<base_domain>`; the SPA prefers the public URL (no VPN gate) and shows public + private URLs side by side.
- **Labs UI** (`70d7bd4`): labs wear their template's attached logo, deployed rows read Running instead of Ready, app chips deduped with a +N overflow.

## Measured result

ERPNext lab, stop → redeploy → Running: **10.7 min before → 111s (1.9 min) after.** Site creation (~100s of real `bench new-site` work) is now the entire cost.

## Verify

Redeploy any Ready lab: the Deploy Log's container step completes in under a second, the assets step does no build, and `https://<instance-id>.benchpress.cloud` (site) and `https://ide-<instance-id>.benchpress.cloud` (code-server) open without the VPN.

Note: the paired infra fix (Traefik attached to the lab docker network, lab network created before compose up) lives in the parent benchpress_devops repo, commit `b3950af`.

Remaining spec work (not in this PR): phase 5 (build/deploy queue split) and phase 4's orphan-detection half.

### DIFF
--- a/benchpress/api.py
+++ b/benchpress/api.py
@@ -7,6 +7,7 @@ from frappe.query_builder import DocType
 from frappe.query_builder.functions import Count
 
 from benchpress import image_cache, lab_detail, lab_templates, labs
+from benchpress.benchpress.doctype.bench_instance.bench_instance import DEPLOY_JOB_TIMEOUT
 from benchpress.credits import account, metering, payments
 from benchpress.credits.guard import (
 	build_charge,
@@ -192,11 +193,7 @@ def create_bench(data: str) -> dict:
 		"benchpress.deploy_manager.deploy_bench",
 		bench_name=doc.name,
 		queue="long",
-		# Covers the measured worst case with headroom: a 7-app first deploy spends
-		# most of an hour in site install + the SSH user's chown over a ~20GB bench.
-		# RQ kills the job at this limit while the real work keeps running orphaned
-		# in the container — the instance then wedges in "Deploying" forever.
-		timeout=7200,
+		timeout=DEPLOY_JOB_TIMEOUT,
 		job_id=f"deploy_bench:{doc.name}",
 		deduplicate=True,
 		enqueue_after_commit=True,
--- a/benchpress/api.py
+++ b/benchpress/api.py
@@ -65,7 +65,7 @@ def build_lab_image(lab_name: str) -> dict:
 		"benchpress.deploy_manager.build_lab",
 		lab_name=lab_name,
 		queue="long",
-		timeout=3600,
+		timeout=10800,
 	)
 	return {"name": lab_name, "status": "Building"}
 
--- a/benchpress/api.py
+++ b/benchpress/api.py
@@ -192,7 +192,11 @@ def create_bench(data: str) -> dict:
 		"benchpress.deploy_manager.deploy_bench",
 		bench_name=doc.name,
 		queue="long",
-		timeout=3600,
+		# Covers the measured worst case with headroom: a 7-app first deploy spends
+		# most of an hour in site install + the SSH user's chown over a ~20GB bench.
+		# RQ kills the job at this limit while the real work keeps running orphaned
+		# in the container — the instance then wedges in "Deploying" forever.
+		timeout=7200,
 		job_id=f"deploy_bench:{doc.name}",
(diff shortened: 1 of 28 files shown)


Venkateshvenki404224:
Phase one: a deploy no longer mounts a fresh named volume at /home/frappe — Docker was eagerly copying the image's entire bench (4-20 GB) into it on every container create. Measured live on the erpnext lab: the container step dropped from ~91s to 0.4s; total deploy time is now dominated by site creation alone.

The same commit fixes two bugs found while proving the phase end-to-end:
- Instance Traefik routes declared `tls: {}` but no certificate in the store covers `base_domain`, so every public instance URL was served Traefik's self-signed default cert and died behind Cloudflare with error 526. Routes now name the letsencrypt resolver with the `*.base_domain` wildcard — issued once, shared by all instances.
- The SPA never received `public_url`: the Lab detail API now returns it, Open buttons prefer it (no VPN gate on a public address), and Connection details shows the public URL and the private VPN URL side by side.

Also raises the lab image build timeout 1h → 3h (a 7-app image legitimately exceeded 1h mid-build).

Verify: redeploy any Ready lab — the Deploy Log's container step completes in under a second, and the instance's https://<id>.<base_domain> URL opens without the VPN. (Paired infra fix in the parent repo attaches Traefik to the lab network; without it instance URLs 504.)

Venkateshvenki404224:
Phase three + phase four (timeouts): the assets step never builds in the container anymore — the dist-dir probe false-positived on SPA-style apps (crm/helpdesk/lms) and asset-less apps (payments/telephony), re-running a 20+ minute `bench build` inside the instance's memory limit on every deploy of such labs. Assets ship in the image (`bench build --production` at image build time). Deploy/redeploy job timeouts raised to 2h to cover the measured worst case (7-app install + chown of a ~20GB bench); the site step now drops a half-created database first, so a killed run can no longer wedge the instance permanently.

Measured after all fixes, ERPNext lab, stop → redeploy → Running: **10.7 min wall** — container step 0.6s, site install ~103s, SSH-user chown ~514s (now the dominant cost), everything else ~1s.

Venkateshvenki404224:
The 10.7-minute redeploy is now **1.9 minutes** (111s click → Running, measured live): linkuser.sh's closing `chown -R` over /home/frappe was a structural no-op — the tenant user is the image's frappe user renamed (uid 1000 unchanged) and every file already belongs to uid 1000; verified zero files needed a change on a running instance. The walk over 120k-1M+ inodes was 514s on the ERPNext lab and ~30 min on the 20GB 7-app lab. The deploy also pushes the app's linkuser.sh into the container before running it, so the fix reaches already-built lab images without a rebuild. SSH user, sudo, write access, and the public URL all verified working after a chown-free deploy.

PR #157, opened 2026-08-24
Implemented phase three

Venkateshvenki404224:
Phase 1 of the `wildcard-cert-routing` spec.

## What this changes

Every bench route file named a certificate resolver, so a bench spawn asked Let's Encrypt for a certificate. Let's Encrypt allows five certificates per identifier set per seven days, so the design had a ceiling of about five benches a week hidden inside it. Nothing issued today only because Traefik de-duplicated the identical request against its store — lose `acme.json`, change `base_domain`, or vary the domain set, and the ceiling becomes real.

Bench routers now carry `tls: {}`: TLS on, no resolver, serving whatever the store already holds for the requested SNI. A single anchor router in `wildcard-anchor.yml` owns `*.<base_domain>` and is the only place in this app that names a resolver. It does not multiply with bench count, so bench churn costs zero ACME calls at any rate.

The two halves ship together on purpose. Bench routers without the anchor is the 526 that reverted this exact shape in PR #150 — `PUBLIC_HOSTNAME` and `BenchPress Settings.base_domain` are different zones, so dropping the resolver alone leaves nothing asking for the bench wildcard.

Details:

- The anchor is written in the `infrastructure` step, minutes ahead of the image build, so a fresh install has room for DNS-01 to finish.
- It rewrites only on a real change, because Traefik reloads on mtime.
- It is never deleted at teardown — it is what keeps the certificate renewing.
- Its `domains` list is fixed, so the resolver is asked for one identity set no matter which SNI arrives. A resolver without `domains` issues per-SNI on demand, which is how a stranger pointing DNS at the host burns the weekly budget.
- The anchor is inert: `priority: 1` (0 is ignored by Traefik) and a deliberately dead backend, so it answers 502.
- No new pipeline step. Still 11.
- `base_domain` unset or `localhost` writes nothing and creates no directory.

## Verified on the live host

A `client-frappe-16` bench deployed after the change, 11/11 steps in 175s:

- Its route file contains `tls: {}` twice and zero `certResolver`; `wildcard-anchor.yml` sits beside it and is the only file in the watched directory naming a resolver.
- `openssl s_client -servername` at the origin returns the Let's Encrypt `benchpress.cloud` / `*.benchpress.cloud` wildcard for both the site and IDE hostnames — not `TRAEFIK DEFAULT CERT`.
- Site answers `{"message":"pong"}` and 200, IDE answers 302, both without `curl -k`.
- `acme.json` holds exactly the two identity sets it held before, byte-identical: the deploy issued nothing.
- Traefik's log shows no ACME, obtain, renew or 429 line for the whole window.
- `tls-anchor.benchpress.cloud` returns 502 on the wildcard, as designed.
- `staging.benchpress.cloud` and the two pre-existing benches keep serving on trusted certificates throughout.

Traefik was never restarted. Every file here is dynamic config in a watched host-bind directory.

## Tests

`run-tests --app benchpress`: 556 tests, and the 13 failures are the same set that fails on a clean `version-16` (credit-ledger, TTL and demo-data timing — none routing-related). `uvx pre-commit@4.3.0 run --all-files` is green.

The spec expected the full-pipeline harnesses to need no change. They did: they never patched the new anchor call, and `/etc/traefik` cannot be created inside the container the tests run in, so every pipeline test failed at step 1. Rather than mock the writers, the harnesses now point `TRAEFIK_DYNAMIC_DIR` at a temporary directory, so they exercise this code instead of skipping it. That also turned the `localhost` case into a real proof that a dev checkout writes nothing at all.

## Known gaps, deferred by the spec

- The deploy log says nothing about the certificate it relies on — phase 2.
- The legacy route files on disk keep their `certResolver` until their bench is next deployed, and the cross-owner misroute the spec documents is untouched — phase 3.
- `BenchPress Settings.on_update` does nothing, so on a fresh install the anchor first appears at the first deploy — phase 3.

### DIFF
--- a/benchpress/benchpress/doctype/benchpress_settings/benchpress_settings.py
+++ b/benchpress/benchpress/doctype/benchpress_settings/benchpress_settings.py
@@ -1,9 +1,31 @@
 # Copyright (c) 2026, Venkatesh and contributors
 # For license information, please see license.txt
 
-# import frappe
+import frappe
 from frappe.model.document import Document
 
 
 class BenchPressSettings(Document):
-	pass
+	def on_update(self):
+		"""Re-anchor the wildcard when the bench zone changes.
+
+		Enqueued, not called: this controller runs in `backend`, which does not mount the
+		Traefik route directory. A direct write would land in that container's own
+		filesystem and Traefik would never see it — silently, with every check green.
+		`queue-long` is the only worker that consumes `long` and the only one with the mount.
+
+		The guard keeps an unrelated settings save — a toggle, a timeout — from queueing a
+		directory sweep.
+		"""
+		if not self.has_value_changed("base_domain"):
+			return
+
+		frappe.enqueue(
+			"benchpress.deploy_manager.reconcile_instance_routes",
+			queue="long",
+			job_id="reconcile_instance_routes",
+			deduplicate=True,
+			# The job re-reads `base_domain` from the database, so it must not start before
+			# the new value is committed. Same mistake `6e25962` fixed for deploys.
+			enqueue_after_commit=True,
+		)
--- a/benchpress/deploy_manager.py
+++ b/benchpress/deploy_manager.py
@@ -222,9 +222,11 @@ def _delete_instance_route(instance_id: str) -> None:
 def reconcile_instance_routes() -> dict:
 	"""Make the Traefik route directory agree with the database.
 
-	Docker and the doctype are the truth; the directory follows. Returns counts rather
-	than a bare success: a reaper that reports "issued" instead of "converged" is how a
-	directory drifts for weeks without anyone noticing.
+	The Bench Instance table is the truth and the directory follows it. Containers are not
+	inspected — an orphaned container is a real problem, but it is todo item 8's, and a
+	pass that quietly took on two jobs would be harder to trust with either. Returns counts
+	rather than a bare success: a reaper that reports "issued" instead of "converged" is how
+	a directory drifts for weeks without anyone noticing.
 
 	Deleting is the load-bearing half. A torn-down container frees its IP back to
 	Docker, which hands it to the next bench, so a route file left behind keeps pointing
--- a/benchpress/deploy_manager.py
+++ b/benchpress/deploy_manager.py
@@ -4,6 +4,8 @@
 import json
 import secrets
 import shlex
+import socket
+import ssl
 from pathlib import Path
 
 import frappe
@@ -73,6 +75,11 @@ TRAEFIK_DYNAMIC_DIR = Path("/etc/traefik/dynamic")
 # certificate resolver. Fixed name so it is idempotently overwritten and can never
 # collide with an instance file, which is always a 32-character hex id.
 WILDCARD_ANCHOR_FILE = "wildcard-anchor.yml"
+
+# Traefik's compose service name. queue-long and traefik both sit on frappe_network, so
+# this resolves; it is the same TLS endpoint the internet reaches, which is the point —
+# checking anything else would prove something else.
+TRAEFIK_HOST = "traefik"
 
 
 def _public_site_url(instance_id: str, base_domain: str | None) -> str | None:
@@ -200,6 +207,58 @@ def _delete_instance_route(instance_id: str) -> None:
 	with that IP next — so this has to run at teardown, not just redeploy.
 	"""
 	(TRAEFIK_DYNAMIC_DIR / f"{instance_id}.yml").unlink(missing_ok=True)
+
+
+def _log_certificate_state(instance_id: str, base_domain: str | None, pipeline) -> None:
+	"""State in the deploy log which certificate this bench's URL will be served on.
+
+	The instance routers name no resolver, so they serve whatever the store already holds.
+	That reliance is otherwise invisible: if the store were ever wrong the first evidence
+	would be a user meeting Cloudflare's 526, which names neither TLS nor a certificate.
+
+	Only the site hostname is checked. The IDE hostname is one label under the same
+	`base_domain` and so is covered by the same wildcard — a second handshake would only
+	re-prove the first one.
+	"""
+	if not base_domain or base_domain == "localhost":
+		return
+
+	hostname = f"{instance_id}.{base_domain}"
+	error = _certificate_error(hostname)
+	if error:
+		pipeline.log(f"WARNING: {error} — the public URL will fail in a browser")
+		return
+
+	pipeline.log(f"TLS ready for {hostname} on the *.{base_domain} wildcard")
+
+
+def _certificate_error(hostname: str, timeout: int = 5) -> str | None:
+	"""Return why `hostname`'s certificate is unusable, or None when it is fine.
+
+	Verifies the way a browser does — full chain and hostname match — against the same TLS
+	endpoint the internet reaches. The store is matched by SNI independently of routing, so
+	this does not race Traefik's file watcher and needs no Traefik API, which this
+	deployment deliberately does not expose.
+
+	Reports rather than raises. A certificate problem must not fail a deploy whose container
+	is up and whose site exists; the owner can still work over the VPN, and the log is where
+	an operator looks.
+	"""
+	context = ssl.create_default_context()
+	try:
+		with socket.create_connection((TRAEFIK_HOST, 443), timeout=timeout) as raw:
+			with context.wrap_socket(raw, server_hostname=hostname):
+				return None
+	except ssl.SSLCertVerificationError as exc:
+		# `verify_message` is set by the C module on a real handshake failure but is absent
+		# on an exception built any other way, and a bare attribute read there would raise
+		# out of the handler — the one thing this function must never do.
+		return f"certificate does not cover {hostname} ({getattr(exc, 'verify_message', None) or exc})"
+	except OSError as exc:
+		# Timeout, refused connection and DNS failure all mean "could not check", not
+		# "certificate is bad". A dev checkout has no Traefik at all, so this is its
+		# normal path — the two causes call for different actions and must stay apart.
+		return f"could not reach Traefik to check {hostname} ({exc})"
 
 
 def _cleanup_failed_deploy(bench, container_id, append_log) -> None:
@@ -391,6 +450,7 @@ def _deploy_bench(bench_name: str) -> None:
 
 		bench.public_url = _public_site_url(bench.name, settings.base_domain)
 		_write_instance_route(bench.name, settings.base_domain, container_ip)
+		_log_certificate_state(bench.name, settings.base_domain, pipeline)
 
 		_setup_container_vpn(bench, container_id, pipeline)
 
--- a/benchpress/deploy_manager.py
+++ b/benchpress/deploy_manager.py
@@ -69,6 +69,11 @@ ADOPTED_MARKER = "already exists — adopting it"
 # control-plane router) lives in this same directory.
 TRAEFIK_DYNAMIC_DIR = Path("/etc/traefik/dynamic")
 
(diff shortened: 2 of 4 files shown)


Venkateshvenki404224:
Live-host follow-up on this phase (no code change — `d778846` stands).

The commit was verified on a **Draft** bench, which proved the generator but never converted a router that was actually serving traffic. Both Running benches still carried `certResolver`. They are now converted **in place**, through this branch's own generator on `queue-long` — the only container that mounts the route directory read-write:

```
docker compose exec -T queue-long bench --site frontend execute \
  benchpress.deploy_manager._write_instance_route \
  --kwargs '{"instance_id":"<id>","base_domain":"benchpress.cloud","container_ip":"<ip>"}'
```

Not `redeploy_bench`: that is `teardown_bench` + `_deploy_bench`, and teardown drops the site database. It would have destroyed two working benches' data for a file that differs only in its `tls` block. Each doc's `container_ip` already matched its live container, so rules, services and backends are byte-identical — only the TLS block changed.

**How to verify**

```bash
# every bench route file: tls: {} twice, no resolver
grep -c certResolver config/traefik/generated/dynamic/<id>.yml     # 0

# the served certificate, at the origin — not through Cloudflare
openssl s_client -connect 127.0.0.1:443 -servername <id>.benchpress.cloud </dev/null 2>/dev/null \
  | openssl x509 -noout -subject -issuer -ext subjectAltName

# nothing was issued
docker exec benchpress_traefik cat /letsencrypt/acme.json | python3 -c \
  'import json,sys; [print(c["domain"]["main"]) for r in json.load(sys.stdin).values() for c in (r.get("Certificates") or [])]'
```

Result: all six bench hostnames (site + `ide-` for three benches) serve `issuer=Let's Encrypt CN=YR2`, SAN `DNS:*.benchpress.cloud, DNS:benchpress.cloud` — no `TRAEFIK DEFAULT CERT`. Through Cloudflare: sites 200, IDEs 302, no 526. `acme.json` identity sets are identical to the pre-change snapshot, and Traefik logged zero `acme`/`obtain`/`renew`/`429` lines — **the deploy and both conversions cost zero ACME calls**, which is the property this PR exists to guarantee. `tls-anchor.benchpress.cloud` returns 502 on the wildcard, inert as designed. `staging.benchpress.cloud` still pongs without `curl -k`.

Tests 556 / 13 failures = the pre-existing baseline set exactly (credit-sweep, TTL, demo-data timing; zero routing). Pre-commit green. Traefik was never restarted — every file here is dynamic config that hot-reloads.

**Still deferred, as the phase spec says:** `091131f5….yml` is an orphan with no Bench Instance doc, so no deploy can convert it; it keeps its resolver until phase 3's route convergence. `dynamic.yml` is the parent repo's control-plane router and is out of scope.

Venkateshvenki404224:
### Phase 1 — live-host convergence complete

The last router still naming a certificate resolver has been converted. Every
bench route file in the watched directory now carries `tls: {}` and no
`certResolver`.

`091131f54bcdfc7bc37cbc45763547fa.yml` is an orphan — no `Bench Instance` doc —
but its URL serves traffic, so it was in scope for this phase: the phase strips
the resolver from the routers *currently serving*. It was rewritten in place
through this branch's own generator on `queue-long`, preserving its backend IP,
so the diff is exactly the two `tls:` blocks. Deleting it and reconciling the
directory to the database stays with phase 3.

**Resolver census, whole directory**

| file | `certResolver` |
|---|---|
| six bench route files | 0 |
| `wildcard-anchor.yml` | 1 — the single anchor |
| `dynamic.yml` | 1 — parent repo's control-plane router, out of scope |

**Verification**

- `acme.json` is **byte-identical** to the snapshot taken before any change.
  Zero ACME calls across the whole phase — the property the feature exists to
  guarantee.
- Traefik's log holds one ACME line in 24h: the daily `Testing certificate
  renew...`. No obtain, no 429.
- All 12 bench hostnames plus the anchor serve `CN=benchpress.cloud`, issuer
  `Let's Encrypt CN=YR2`, SAN `DNS:*.benchpress.cloud, DNS:benchpress.cloud` at
  the origin. No `TRAEFIK DEFAULT CERT`.
- Sites 200 and IDEs 302 both at the origin without `-k` and through Cloudflare.
  `staging.benchpress.cloud` pongs without `-k`.
- `tls-anchor.benchpress.cloud` → 502 on the wildcard. Inert as designed.
- 556 tests, 13 failures — the pre-existing baseline set exactly (credit-sweep,
  TTL, demo-data, signup; zero routing). All 11 routing/anchor tests pass.
  `pre-commit` green.
- Traefik was never restarted or recreated; every change hot-reloaded from the
  watched directory.

No code changed for this — it was live routing state, so `d778846` remains the
phase-1 commit.

**Reproduce**

```bash
grep -c certResolver config/traefik/generated/dynamic/*.yml
openssl s_client -connect 127.0.0.1:443 -servername <id>.benchpress.cloud </dev/null 2>/dev/null \
  | openssl x509 -noout -subject -issuer -ext subjectAltName
```

Venkateshvenki404224:
## Phase 2 — certificate verification in the deploy log

Phase 1 made bench routers name no certificate resolver, so they serve whatever
Traefik's store already holds. Nothing stated that the store was right. This
phase makes that reliance visible at deploy time instead of leaving it to a user
meeting Cloudflare's 526 — an error that names neither TLS nor a certificate.

### What it adds

- `_certificate_error(hostname)` completes a TLS handshake against Traefik and
  verifies the way a browser does, full chain and hostname match. It returns a
  string or `None`; it never raises. Two distinct messages, because the causes
  call for different actions: `certificate does not cover <host>` means fix the
  certificate, `could not reach Traefik to check <host>` means Traefik is down —
  or that there is no Traefik, which is the dev-checkout case.
- `_log_certificate_state(...)` writes the outcome into the deploy log during
  step 4 of 11. No new pipeline step; the eleven-step contract is unchanged.
- `TRAEFIK_HOST = "traefik"`, the same TLS endpoint the internet reaches.

A TLS handshake rather than a router poll: this deployment omits Traefik's API
section on purpose, and the handshake tests the thing that actually broke. The
store is matched by SNI independently of routing, so it does not race the file
watcher.

**Non-fatal by design.** A certificate problem must not fail a deploy whose
container is up and whose site exists — the owner can still work over the VPN.

**This phase adds no router and no `tls.domains` block, so it requests no
certificate.** `acme.json` is byte-identical across the whole phase.

### How to verify

1. `bench --site frontend run-tests --app benchpress` and
   `uvx pre-commit@4.3.0 run --all-files`.
2. Deploy a bench and read step 4 of its Deploy Log:
   ```
   TLS ready for <id>.benchpress.cloud on the *.benchpress.cloud wildcard
   ```
3. The report-never-raise contract, from the worker that runs deploys:
   ```bash
   docker compose exec queue-long bench --site frontend console
   ```
   ```python
   from benchpress.deploy_manager import _certificate_error
   _certificate_error("nope.example.com")       # a string, naming the host
   _certificate_error("<id>.benchpress.cloud")  # None
   ```
4. Confirm nothing was issued:
   ```bash
   docker exec benchpress_traefik md5sum /letsencrypt/acme.json
   ```

### Verified on staging

- Tests: 13 failures, the same set as the `version-16` baseline (credit sweep,
  TTL, demo-data — all timing, none routing). 11 new tests pass. Lint green.
- A real deploy logged
  `TLS ready for ad42c0a5….benchpress.cloud on the *.benchpress.cloud wildcard`
  inside `Step 4/11`.
- On `queue-long`: `nope.example.com` returned
  `certificate does not cover nope.example.com (self-signed certificate)`, a live
  bench hostname returned `None`, neither raised.
- `acme.json` md5 unchanged across the phase; still exactly two identity sets;
  no ACME, obtain, renew or 429 lines in the Traefik log.
- All bench URLs and the control plane still serve 200 at the origin on the
  Let's Encrypt wildcard. Traefik was never restarted.

### Notes

- 11 new tests. None opens a socket — `_certificate_error` is patched and the
  assertions are on the log, so no test depends on a running Traefik.
- The four full-deploy harnesses now patch `_certificate_error`, which would
  otherwise open a real socket during a unit test.
- `verify_message` is absent on an `SSLCertVerificationError` built any way other
  than a real handshake, so it is read with `getattr` — a bare read would raise
  out of the handler, the one thing this function must never do.

Venkateshvenki404224:
## Phase 3 — route-file convergence

Closes the wildcard-cert-routing spec. Phase 1 changed what a route file says, phase 2 made it measurable, and phase 3 makes the *set* of files converge instead of accrete.

### What this adds

- `deploy_manager.reconcile_instance_routes()` — anchors the wildcard, writes a route for every `Running` bench that has a `container_ip`, and deletes every other instance file. Returns `{"anchored", "written", "deleted", "kept"}`.
- `deploy_manager._routable_instance_ips()` — the `frappe.qb` read that decides which benches own a route.
- `PROTECTED_ROUTE_FILES` — `dynamic.yml` (the parent repo's control-plane router) and `wildcard-anchor.yml` (this feature's certificate) are skipped by name, never by pattern. Deleting either takes the platform off the internet.
- `BenchPress Settings.on_update` — a changed `base_domain` enqueues the pass on `queue="long"` with `enqueue_after_commit=True`. That worker is the only one that mounts the route directory, and the job re-reads the domain, so it must not start before the commit.
- 10 tests: 8 for the pass, 2 for the trigger.

### The live misroute this fixed

Three route files named no Bench Instance at all. One of them, `091131f5…`, pointed at 172.30.0.5, which Docker had since handed to `9814cb07…` — a bench owned by another account. Before the pass, `https://091131f5….benchpress.cloud/api/method/ping` returned `200 {"message":"pong"}` from that bench. Two more pointed at 172.30.0.4, which is now Traefik itself.

### How to verify

```bash
docker compose exec queue-long bench --site frontend execute \
  benchpress.deploy_manager.reconcile_instance_routes
```

Measured on staging:

| check | result |
|---|---|
| first run | `{"anchored": false, "written": 3, "deleted": 3, "kept": 2}` |
| second run (idempotence) | `{"anchored": false, "written": 3, "deleted": 0, "kept": 2}` |
| directory after | `dynamic.yml`, `wildcard-anchor.yml`, one file per running bench, each at its current IP |
| the three orphan hostnames | `404` — were `200`, `502`, `502` |
| the three running benches | `200 {"message":"pong"}` and `302` on the IDE, no `-k` |
| certificate at the origin | `CN=benchpress.cloud`, Let's Encrypt YR2, SAN `*.benchpress.cloud, benchpress.cloud` |
| `staging.benchpress.cloud/api/method/ping` | `pong`, no `-k` |
| `base_domain` changed, then changed back | anchor and all three route files followed both ways, with no deploy |
| `base_domain` saved unchanged | nothing enqueued, no file mtime moved |
| `acme.json` | md5 `ae3d5562361983c041fb2e0cba16ee31` before and after, two identity sets, zero ACME log lines |
| Traefik | never restarted, `RestartCount=0` |

Tests: 569 run, 13 failures, the same demo-data and credit-sweep set that fails on `version-16`. `uvx pre-commit@4.3.0 run --all-files` green.

Venkateshvenki404224:
### One consequence to know about before merging

`enqueue_start` returns a `Stopped` bench to `Running` but never writes a route file, and `stop_bench` deliberately keeps the document. So a sweep that lands between a stop and a start now deletes that bench's route, and the bench comes back with no route until the next deploy or sweep. Its URL returns 404 in that window.

Before this phase it kept the old file instead — pointing at an IP Docker may already have given to somebody else. The 404 is the safer of the two, and "no file for a bench that is not running" is this phase's stated goal, so this is the intended trade rather than a gap.

The real fix is routing that follows `stop` / `start` / `restart`, scoped in `specs/not-completed/restart-free-dynamic-routing/` (todo item 2), which is blocked on this branch merging because both change `_write_instance_route`.

PR #158, opened 2026-08-24
Implemented phase three

Venkateshvenki404224:
## Phase 1 — Name-based route files

Traefik route files named the bench's container IP on a recycled `/24`, and no
lifecycle transition rewrote them. A stopped or redeployed bench left its hostname
pointing at an address Docker had already handed to another owner's bench.

`_write_instance_route` now takes `(instance_id, base_domain)` and points both the
site and IDE services at the **container name**, which Docker's embedded DNS
resolves (Traefik sits on the `benchpress` network). The route file is a pure
function of its two arguments, so nothing between spawn and teardown can make it
stale. A recycled address can no longer misroute; a removed container gives a 502,
which is the honest answer.

Route files are also written through `_atomic_write` (temp file + `os.replace`), so
the file provider never reads a half-written config. The wildcard anchor uses it too.

### Verified on the live staging host

- Redeployed a bench: both services name the bench id, no address anywhere in the
  file, `tls: {}` still present with no `certResolver`.
- All three bench hostnames serve `{"message":"pong"}` at the origin with **no `-k`**,
  each answering from its own server (distinct Python versions), not the control plane.
- **The measurement:** `docker stop` + `docker start` left the route file's mtime and
  md5 **identical**. First curl at +0.03 s returned 502, recovering at ~1.7 s — the
  container's web server binding its port, not a routing fault: the container's address
  never changed across the cycle, so stale-pool staleness cannot explain it.
- Traefik `StartedAt` unchanged (`2026-08-20T05:26:55Z`, `RestartCount=0`), zero errors
  in its log, zero ACME activity. `acme.json` md5 unchanged, still exactly 2 identity sets.
- `staging.benchpress.cloud` still serves the control plane.
- `reconcile_instance_routes` converged the two remaining IP-based files to names
  (`written: 3, deleted: 0, kept: 2`); both protected files untouched.

### Tests

`test_deploy_manager`: 67 integration + 24 unit, all green, including three new tests
(container-named backends with no dotted quad anywhere, no `.tmp` left after a rewrite,
no `.tmp` left by the anchor). Full suite has 13 pre-existing environmental failures in
`test_credit_sweep` / `test_demo_data` — confirmed identical with this change stashed out.
`uvx pre-commit@4.3.0 run --all-files`: all 12 hooks pass.

### DIFF
--- a/benchpress/api.py
+++ b/benchpress/api.py
@@ -223,6 +223,7 @@ def _assert_site_name_changeable(doc) -> None:
 
 @frappe.whitelist()
 def bench_action(bench_name: str, action: str) -> dict:
+	from benchpress.deploy_manager import enqueue_route_sync
 	from benchpress.docker_manager import restart_container, start_container
 
 	require_bench_access(bench_name)
@@ -252,6 +253,9 @@ def bench_action(bench_name: str, action: str) -> dict:
 	metering.on_bench_running(bench)
 	bench.save()
 	frappe.db.commit()
+	# A restart re-allocates the container address, and a start may be following a stop that
+	# removed the file — both need the route re-established.
+	enqueue_route_sync(bench.name)
 	return {"name": bench.name, "status": bench.status}
 
 
--- a/benchpress/benchpress/doctype/bench_instance/bench_instance.py
+++ b/benchpress/benchpress/doctype/bench_instance/bench_instance.py
@@ -93,6 +93,7 @@ class BenchInstance(Document):
 	@requires_credits(cost=instance_runway, caps=(cap_concurrent_instances,))
 	def enqueue_start(self):
 		from benchpress.credits import metering
+		from benchpress.deploy_manager import enqueue_route_sync
 		from benchpress.docker_manager import start_container
 
 		if not self.container_id:
@@ -103,4 +104,5 @@ class BenchInstance(Document):
 		metering.on_bench_running(self)
 		self.save()
 		frappe.db.commit()  # nosemgrep: intentional commit to persist status before response
+		enqueue_route_sync(self.name)
 		frappe.msgprint(_("Bench started."))
--- a/benchpress/deploy_manager.py
+++ b/benchpress/deploy_manager.py
@@ -65,11 +65,9 @@ SITE_HTTP_PORT = 8000
 ADOPTED_MARKER = "already exists — adopting it"
 
 # Module constant, not inlined, so tests can monkeypatch it to a tmp path.
-# Flat, not a subdirectory: Traefik's file provider does not recurse, so this
-# must be the exact directory the `directory:` provider in traefik.yml.template
-# watches, and the exact host path bind-mounted read-write into queue-long /
-# read-only into traefik in docker-compose.prod.yml. dynamic.yml (the
-# control-plane router) lives in this same directory.
+# Flat, not a subdirectory: Traefik's file provider does not recurse, so this must be the
+# exact directory the `directory:` provider in traefik.yml.template watches. dynamic.yml
+# (the control-plane router) lives in this same directory.
 TRAEFIK_DYNAMIC_DIR = Path("/etc/traefik/dynamic")
 
 # One file, one router, one identity set — the only place in this app that names a
@@ -91,6 +89,15 @@ PROTECTED_ROUTE_FILES = frozenset({CONTROL_PLANE_ROUTE_FILE, WILDCARD_ANCHOR_FIL
 # this resolves; it is the same TLS endpoint the internet reaches, which is the point —
 # checking anything else would prove something else.
 TRAEFIK_HOST = "traefik"
+
+
+class TraefikRouteDirectoryMissing(Exception):
+	"""Raised when the Traefik route directory is not mounted in this container.
+
+	`queue-long` mounts it read-write and traefik read-only; `backend` and
+	`queue-short` mount neither, and both consume `default`. Every caller therefore
+	reaches routing through `enqueue_route_sync`, which pins `queue="long"`.
+	"""
 
 
 def _public_site_url(instance_id: str, base_domain: str | None) -> str | None:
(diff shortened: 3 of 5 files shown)


Venkateshvenki404224:
## Phase 2 — status-driven route sync

A route file now exists exactly while its bench is running.

**What changed**

- `deploy_manager.sync_instance_route(bench_name)` — the one decision point. Writes when `status == "Running"`, deletes otherwise, returns `written` / `deleted` / `skipped`. A bench that no longer exists reads as no status, which deletes rather than raises. `base_domain` of `localhost` or unset skips entirely, so a dev checkout is untouched.
- `deploy_manager.enqueue_route_sync(bench_name)` — the only way callers reach it. Pins `queue="long"`, the one worker that mounts the route directory, with `job_id=route_sync:<name>`, `deduplicate=True` and `enqueue_after_commit=True`.
- `stop_bench`, `api.bench_action` (`start` and `restart`) and `BenchInstance.enqueue_start` each enqueue one sync after their commit.
- `teardown_bench` still deletes directly — it already runs on `queue-long`, and a teardown must not depend on a second job surviving.
- `stats_collector._stop_if_dead` needed no call: it stops through `stop_bench`, which now syncs.

**How to verify**

Against the origin rather than through Cloudflare, so a failure means the origin is broken:

```bash
B=<bench-id>; H=$B.benchpress.cloud
D=config/traefik/generated/dynamic

# stop from the SPA, then:
ls $D                                    # $B.yml is gone
curl -sk -o /dev/null -w '%{http_code}\n' --resolve "$H:443:127.0.0.1" "https://$H/"
# 404 from Traefik — a name with no router, not another owner's bench

# start from the SPA, then:
ls $D                                    # $B.yml is back
curl -sS -i --resolve "$H:443:127.0.0.1" "https://$H/api/method/ping" | grep -i '^server:'
# must NOT be nginx — nginx means the control plane answered a bench hostname
```

The queue is the correctness property, so check it explicitly. Pause the worker to catch the job queued:

```bash
docker compose stop queue-long
# start the bench from the SPA, then:
docker compose exec -T backend bench --site frontend console
>>> from frappe.utils.background_jobs import get_queue
>>> [j.id for j in get_queue("long").jobs]
docker compose start queue-long
```

**Measured on staging**

| Check | Result |
|---|---|
| Stop removes the file | gone in ~750 ms; both `<id>` and `ide-<id>` return Traefik's bare 404 |
| Start restores it | back in ~1.25 s; `server: Werkzeug/3.1.6 Python/3.14.2`, `pong`, no `-k` |
| Restart re-syncs | file rewritten; serves after ~2 s of container boot |
| Queue | `frontend||route_sync|<id>` queued on **long**; 0 on `default`, 0 on `short`. Every `RQ Job` row for `sync_instance_route` reads `queue: long`, `status: finished` |
| Untouched benches | both route files' mtimes unchanged; both still answer `pong` from their own container |
| Control plane | `pong`, `server: nginx/1.22.1`, no `-k` |
| Traefik | `StartedAt` unchanged, `RestartCount 0`, 0 errors or `429` in the last hour |
| Certificates | `acme.json` md5 byte-identical to before |

Tests: `test_deploy_manager` 67 integration + 24 unit, all OK, including 6 new `sync_instance_route` cases and 6 trigger cases that assert `queue="long"` and `enqueue_after_commit=True` by name. Full suite `Ran 581, FAILED (failures=13)` — the same 13 pre-existing failures in `test_credit_sweep` and `test_demo_data` that phase 1 recorded, none in `test_deploy_manager`. `uvx pre-commit@4.3.0 run --all-files` green.

One side effect worth knowing: `test_credit_sweep` calls the real `stop_bench`, so a test run now enqueues real `route_sync` jobs on the live host. They are no-ops — the test bench rows are gone by the time the worker reads them, so the sync deletes a file that was never written.

Venkateshvenki404224:
## Phase 3 — enforce the mount invariant, and let drift self-heal

Closes the spec. Phase 1 changed what a route file says, phase 2 made the set of files
follow each bench's status, and this phase makes both hard to break.

### What it adds

**A routing write from a container that cannot reach Traefik now fails loudly.**
`_write_instance_route` and `_ensure_wildcard_anchor` used to call
`TRAEFIK_DYNAMIC_DIR.mkdir(parents=True, exist_ok=True)`. That turned "this process
cannot reach Traefik" into "this process wrote a file into its own filesystem" — the same
outcome with none of the evidence. Both calls are gone. `_atomic_write` raises
`TraefikRouteDirectoryMissing` instead, so every writer inherits the guard from the one
place that writes. The message names the path, the bind mount and `enqueue_route_sync`,
because a bare `FileNotFoundError` leaves the reader to rediscover the mount topology.
That topology now lives in the exception's own docstring and nowhere else — the three
comments that used to repeat it are deleted.

**The route directory converges every five minutes.** `hooks.py` gains
`benchpress.deploy_manager.enqueue_route_reconcile` on the `*/5` cron — the enqueuer,
never `reconcile_instance_routes` itself. Frappe sends scheduled jobs to `default`, which
`queue-short` also consumes, and that container has no mount; scheduling the pass directly
would raise every five minutes. The enqueuer pins `queue="long"` and deduplicates on a
fixed `job_id`.

**One change beyond the phase text.** `_atomic_write` now skips an identical write and
returns whether it wrote. Without it a `*/5` pass rewrites every running bench's route
file every tick, and Traefik reloads on mtime — a reload every five minutes to say
nothing, in a feature named for not needing one. `_ensure_wildcard_anchor` had its own
copy of that check and now composes `_atomic_write` instead.

Tests arrange the mount the way production has it (new `_mounted(tmp)` helper, eight call
sites) rather than being exempt from the guard. The two dev-checkout cases deliberately
stay unmounted — that is now what proves the `localhost` return happens before the guard.

### How to verify

```bash
uvx pre-commit@4.3.0 run --all-files

# The cron row only exists after jobs are synced:
docker compose exec -T backend bench --site frontend execute \
  frappe.core.doctype.scheduled_job_type.scheduled_job_type.sync_jobs

# The guard fires where there is no mount. Use the console, not `bench execute` —
# `bench execute` catches the exception and re-raises a NameError over it.
docker compose exec -T backend bench --site frontend console <<'PY'
from benchpress.deploy_manager import sync_instance_route
try:
    sync_instance_route("<running-id>")
except Exception as e:
    print(type(e).__name__, e)

PY
docker compose exec -T backend ls /etc/traefik    # must still not exist

# And not where there is one:
docker compose exec -T queue-long bench --site frontend execute \
  benchpress.deploy_manager.sync_instance_route --kwargs '{"bench_name":"<running-id>"}'

# Drift heals. Plant a copy of a real route file so a parse error cannot be mistaken
# for convergence, then wait one tick:
cp config/traefik/generated/dynamic/<running-id>.yml \
   config/traefik/generated/dynamic/deadbeefdeadbeefdeadbeefdeadbeef.yml

# The pass runs only on queue-long:
docker compose logs queue-short --since 15m | grep -i reconcile_instance_routes  # nothing
docker compose logs queue-long  --since 15m | grep -i reconcile_instance_routes  # the runs
```

### Measured on staging

- The guard raised in `backend` with the message naming `queue-long`; `/etc/traefik` still
  does not exist there. The same call in `queue-long` returned `written`, and the route
  file's md5 and mtime were both unchanged.
- A planted stale route file, a byte copy of a live one, was removed **within one tick**
  (15:44:07 → 15:48:14 UTC). Traefik had parsed it — its log shows `HTTP router already
  configured, skipping`, a duplicate-name warning rather than a parse error — so the
  removal is convergence. `dynamic.yml`, `wildcard-anchor.yml` and all three running
  benches' files survived.
- The tick split as designed: `queue-short` ran only the thin enqueuer on `default`;
  `queue-long` ran the pass as `(frontend||route_reconcile) … Job OK`. Grepping any
  container log for `TraefikRouteDirectoryMissing` finds nothing.
- **Three ticks across a fifteen-minute idle window changed no file's mtime.**
- Control plane serves `pong` over HTTP/2 with no `-k`; all three benches serve from their
  own gunicorn. No route file contains an IPv4 address.
- Traefik `StartedAt` unchanged and `RestartCount 0` throughout; `acme.json` md5 identical
  to phases 1 and 2, still two identity sets. Nothing was issued, and Traefik was never
  restarted, recreated or `up`ed.

One pre-existing failure set remains: 13 tests in `test_credit_sweep` (11) and
`test_demo_data` (2), the same ones phases 1 and 2 recorded. Every one is an assertion
comparing bench-id lists — live bench rows leaking into a sweep assertion, unrelated to
routing.

PR #159, opened 2026-08-24
Implemented phase three

Venkateshvenki404224:
Phase 1 of the `sysbox-default-runtime` spec: a bench names the runtime it runs on, and `sysbox` is available as a choice.

## What this adds

- `CONTAINER_RUNTIMES` in `docker_manager.py` — an allow-list mapping a field value to a runtime name the daemon has registered. `None` means "pass no runtime", which leaves the daemon's `default-runtime`.
- `resolve_runtime(bench_doc)` — throws on anything outside the allow-list, so a stored string never reaches `containers.create` unchecked.
- `runtime` Select on Bench Instance (`runc` / `sysbox`), default `runc`.
- The `runtime=` kwarg is passed **only** when the resolved value is not `None`, so a bench on `runc` produces the same `containers.create` call it produced before this branch.

The host half is in `benchpress_devops` on the branch of the same name: `scripts/enable-sysbox.sh` and a `PRODUCTION.md` §5.17 findings entry.

## How to verify

Deployed a real bench (`crm` lab) with `runtime = sysbox` on staging:

```
docker inspect --format '{{.HostConfig.Runtime}}' <c>   -> sysbox-runc
docker exec <c> cat /proc/self/uid_map                  -> 0 231072 65536
docker exec <c> id                                      -> uid=0(root)
```

`uid_map` is the check that matters — `HostConfig.Runtime` records what was asked for. The identity map `0 0 4294967295` (what every bench read before this change) would mean Sysbox did nothing. Container root is now host UID 231072, unprivileged.

WireGuard still works under the user namespace — `wg show` lists the peer with a live handshake, `wg0` holds its `172.27.0.8/32`. The site serves: `HTTP 200`, `/api/method/ping` returns `pong`. Deploy ran all 11 steps in 46.0s; the container-create step cost 0.4s.

The three pre-existing benches still run on `runc` and still read the identity `uid_map` — unchanged. `docker info` still reports `Default Runtime: runc`.

Backend tests: the three new runtime cases pass. Frontend: 154 pass. `uvx pre-commit@4.3.0 run --all-files`: 12 hooks pass.

## Deliberately not in this phase

Sysbox is not the default, the field is not admin-only, existing rows carry no backfilled value, and a bench asking for a runtime the host lacks still fails late with Docker's own error. Phases 2 and 3.

### DIFF
--- a/benchpress/api.py
+++ b/benchpress/api.py
@@ -97,6 +97,7 @@ def get_benches() -> list[dict]:
 			"status",
 			"container_id",
 			"container_ip",
+			"runtime",
 			"wg_ip",
 			"cpu_usage",
 			"memory_usage",
--- a/benchpress/api.py
+++ b/benchpress/api.py
@@ -525,3 +525,12 @@ def run_diagnostics() -> list[dict]:
 	from benchpress.diagnostics import run_diagnostics as _run_diagnostics
 
 	return _run_diagnostics()
+
+
+@frappe.whitelist()
+def preflight_runtime(runtime: str) -> dict:
+	"""Prove a runtime works, which `run_diagnostics` cannot: this one creates a container."""
+	require_admin()
+	from benchpress.docker_manager import preflight_runtime as _preflight_runtime
+
+	return _preflight_runtime(runtime)
--- a/benchpress/benchpress/doctype/bench_instance/bench_instance.json
+++ b/benchpress/benchpress/doctype/bench_instance/bench_instance.json
@@ -24,6 +24,7 @@
   "container_section",
   "container_id",
   "container_image",
+  "runtime",
   "column_break_container",
   "container_ip",
   "started_at",
@@ -143,6 +144,14 @@
    "fieldtype": "Data",
    "label": "Container Image",
    "read_only": 1
+  },
+  {
+   "default": "runc",
+   "description": "Container runtime. sysbox user-namespaces the container so container root is unprivileged on the host.",
+   "fieldname": "runtime",
+   "fieldtype": "Select",
+   "label": "Runtime",
+   "options": "runc\nsysbox"
   },
   {
    "fieldname": "column_break_container",
@@ -296,7 +305,7 @@
    "link_fieldname": "bench"
   }
  ],
- "modified": "2026-08-19 00:00:00.000000",
+ "modified": "2026-08-24 18:00:00.000000",
  "modified_by": "Administrator",
  "module": "Benchpress",
  "name": "Bench Instance",
--- a/benchpress/benchpress/doctype/bench_instance/bench_instance.json
+++ b/benchpress/benchpress/doctype/bench_instance/bench_instance.json
@@ -146,12 +146,12 @@
    "read_only": 1
   },
   {
-   "default": "runc",
    "description": "Container runtime. sysbox user-namespaces the container so container root is unprivileged on the host.",
    "fieldname": "runtime",
    "fieldtype": "Select",
    "label": "Runtime",
-   "options": "runc\nsysbox"
+   "options": "\nrunc\nsysbox",
+   "permlevel": 1
   },
   {
    "fieldname": "column_break_container",
@@ -305,7 +305,7 @@
    "link_fieldname": "bench"
   }
  ],
- "modified": "2026-08-24 18:00:00.000000",
+ "modified": "2026-08-24 19:30:00.000000",
  "modified_by": "Administrator",
  "module": "Benchpress",
  "name": "Bench Instance",
@@ -342,6 +342,23 @@
    "report": 1,
    "role": "BenchPress User",
    "write": 1
+  },
+  {
+   "permlevel": 1,
+   "read": 1,
+   "role": "System Manager",
+   "write": 1
+  },
+  {
+   "permlevel": 1,
+   "read": 1,
+   "role": "BenchPress Admin",
+   "write": 1
+  },
+  {
+   "permlevel": 1,
+   "read": 1,
+   "role": "BenchPress User"
   }
  ],
  "row_format": "Dynamic",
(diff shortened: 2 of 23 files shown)


Venkateshvenki404224:
## Phase 2 — Sysbox by default, admin-only to change

A bench with nothing set takes its runtime from `BenchPress Settings.default_bench_runtime`, seeded to `sysbox`. A tenant cannot change it through any API path, but can still see what they got.

**What landed**

- `BenchPress Settings.default_bench_runtime` — Select, next to `default_image`. An install on a host without Sysbox opts out here instead of carrying a patch.
- `Bench Instance.runtime` gains `permlevel: 1`. System Manager and BenchPress Admin read + write it; BenchPress User reads only.
- `patches/backfill_bench_runtime.py` — stamps `runc` on every bench that predates the field, **then** seeds `sysbox`. Rows that already name a runtime are left alone, so phase 1's Sysbox bench is not relabelled either way.
- The controller takes the Settings default on insert and refuses a change unless the bench is `Draft`. `Stopped` is refused with `Running`: a stopped bench still owns a container built under the old runtime.

**Two framework behaviours the code has to work around** — both are why the shape differs from the phase spec:

- Frappe resets a permlevel field the caller cannot write back to its stored value and raises nothing, so a tenant lowering their own isolation would read as success. `validate_higher_perm_levels` is overridden to turn that silent reset into a `PermissionError`.
- Frappe treats a Select's first option as the value when there is no default, so `runtime` would have taken `runc` before Settings was ever read. The Select carries an empty first option, and the default is chosen in `validate` rather than `before_insert` — the permlevel reset runs between the two.

**How to verify**

```bash
docker compose exec backend bench --site frontend migrate
# every bench that predates this reads runc; Settings reads sysbox
```

Verified on staging:

- A `BenchPress User` deployed through `create_bench` setting nothing → 11/11 steps in 45.5s, `HostConfig.Runtime` = `sysbox-runc`, `/proc/self/uid_map` = `0 231072 65536` (container root is an unprivileged host UID), site returns 200 and `/api/method/ping` → `pong`.
- That tenant calling `frappe.client.set_value(runtime="runc")` on their own bench → `PermissionError`; field still reads `sysbox`. `get_benches` still returns `runtime` to them.
- An admin changing `runtime` on a `Running` bench and on a `Stopped` one → refused, message naming the instance. On a `Draft` bench → allowed.
- 596 backend tests, 13 failures, all pre-existing (11 `test_credit_sweep`, 1 `test_demo_data`, 1 `test_signup`). 154 frontend tests pass. `pre-commit` clean.
- The three benches that predate this branch are untouched: still `running`, still `runc`, still the identity uid_map. Daemon `default-runtime` is still `runc`.

**Rollback:** set `BenchPress Settings.default_bench_runtime = runc`. One field, no deploy. Existing Sysbox containers keep running.

Still missing, and phase 3's job: a bench asking for a runtime the daemon lacks still fails at `containers.create`, minutes in, with Docker's own message.

Venkateshvenki404224:
## Phase 3 — fail fast, and say what actually ran

Phase 2 made Sysbox the default, which makes the host a single point of failure for every deploy. This phase keeps that failure loud and fast.

**What it adds**

- `docker_manager.host_runtimes()` — the daemon's registered runtimes and its default, one `docker info` memoised per job through `request_cache.local_cache`.
- `docker_manager.preflight_runtime(name)` — runs a throwaway `alpine` container under the runtime and returns `{"ok", "detail"}`. It reports instead of raising, and it is the only check that proves a runtime *works*: `sysbox-runc` stays listed in `docker info` with its unit active while unable to create a container.
- `deploy_manager._assert_runtime_registered` — the gate, placed **before** `pipeline.step("image")`. No fallback to `runc`: a bench that cannot be isolated fails and an admin decides.
- One log line per deploy, read back off the container: `container runtime sysbox-runc`.
- A sixth diagnostics row, `container_runtimes`. `tests/test_api.py`'s fixture, `overview.py`'s docstring and `InfrastructureCard.vue`'s comment moved with it.
- `api.preflight_runtime`, admin-only beside `run_diagnostics`.
- `runtime` renders read-only in `ConnectionDetails.vue` and the Instances table (`data/benches.js` needed the field too — that list resource has its own field list).

**How to verify**

```bash
# the gate, timed — inside queue-long, which owns the traefik mount
docker compose exec -T queue-long bench --site frontend console
# patch docker_manager._read_host_runtimes to a set without sysbox-runc, then
# deploy_manager.deploy_bench(<a Draft bench>)

# the negative control and the verdict line
docker compose exec -T backend bench --site frontend console
# api.create_bench('{"lab": "crm"}') as a BenchPress User, then read its Deploy Log
docker inspect -f '{{.HostConfig.Runtime}}' <container>

# preflight through the API
docker compose exec -T backend bench --site frontend execute benchpress.api.preflight_runtime --kwargs '{"runtime":"sysbox"}'
```

**Measured on staging**

- Unavailable runtime: refused in **0.4 s**, `Step 2/11` never written, message `Bench runtime 'sysbox-runc' is not registered with the Docker daemon (it has: io.containerd.runc.v2, runc).`
- Negative control: a `BenchPress User` deploy of `crm` reached 11/11 in 43.5 s; the log carries `container runtime sysbox-runc`, `docker inspect` agrees, `/proc/self/uid_map` reads `0 231072 65536`, the site answers 200 and `pong`.
- `preflight_runtime`: `{"ok": true, "detail": "sysbox-runc ran a container"}` for an admin in 0.8 s; `PermissionError` for a `BenchPress User`; `{"ok": false}` carrying Docker's own message for a runtime the daemon rejects.
- Admin infrastructure screen: six rows, the sixth `Container runtimes — Active`.
- 601 tests, the same 13 pre-existing failures (11 `test_credit_sweep`, 1 `test_demo_data`, 1 `test_signup`); frontend 154 pass; `pre-commit` clean.

The "registered but broken" case is covered by a unit test asserting Docker's message passes through verbatim rather than by `systemctl stop sysbox` — two tenant benches are running under `sysbox-runc` on this host and stopping `sysbox-fs` would degrade them.

PR #160, opened 2026-08-24
Implemented phase five

Venkateshvenki404224:
Phase 1 of the [lease-based-credits](specs/in-progress/lease-based-credits/README.md) spec.

This phase produces numbers, not features. It ships one guard — the clock-skew
diagnostic — and measures the two unknowns the rest of the spec is priced against.

## What ships

`benchpress/diagnostics.py` gains a `clock_skew` check. Frappe writes every Datetime
naive in the site timezone; SQL `NOW()` answers in the database session's. When the two
differ, any deadline the database evaluates is wrong by that difference — and it reads
as a lease that has simply not expired yet.

The check is **red on this host**, which is the point:

```
App clock says 2026-08-25 04:45:05 and the database says 2026-08-24 23:15:05
— a gap of 19800s. Compare stored deadlines against an epoch integer bound in
Python, never against SQL NOW().
```

Tolerance is a constant, not a setting: 2 seconds, which absorbs MariaDB truncating
`NOW()` to whole seconds and nothing else. The database clock is read with
`frappe.qb.select(Now())`, so the app still has zero `frappe.db.sql` calls.

The row count is stated in three places, not the two the spec names — the `test_api`
fixture, the `overview.py` docstring, and a comment in `InfrastructureCard.vue`. All
three moved.

## Measurement 1 — the drain rate

Against a real bench container deployed for this test, fully loaded (11 processes),
20 s settle before each stop. Not `alpine`.

| Command | n | p50 | max | exit |
|---|---|---|---|---|
| `docker stop -t 30` (what `stop_container` does) | 5 | **30.556 s** | 30.583 s | **137** |
| `docker stop` (default 10) | 5 | 10.604 s | 10.628 s | **137** |
| `docker stop -t 5` | 5 | 5.593 s | 5.607 s | **137** |
| `docker kill` (the floor) | 3 | 0.579 s | 0.580 s | **137** |

`deploy_manager.stop_bench()` end to end: **30.916 s**. All the Frappe work costs
**0.36 s — 1.2% of the job**.

**Every stop is a SIGKILL.** Elapsed is exactly `grace_period + 0.58 s` in all four
rows, and the exit code is 137 in all 18 trials. PID 1 in a bench container is `tail`,
which installs no SIGTERM handler, and the kernel discards default-action signals sent
to PID 1 of a namespace. SIGTERM is dropped; Docker always falls through to SIGKILL
after the full timeout.

Drain of 100 simultaneous expiries, one `queue-long` worker, serialized:

| Stop mode | 100 benches, 1 worker | workers for < 60 s |
|---|---|---|
| `-t 30` — today | **50.9 min** | **51** |
| `-t 5` | 9.3 min | 10 |
| `kill` / no grace | **58 s** | **1** |

The spec's worst case — "a hundred-bench burst takes fifty minutes" — is not a worst
case. It is the measured median.

**This re-scopes phase 4.** The ceiling is PID 1, not the worker count. Give the
container an init that handles SIGTERM and one worker nearly suffices; add parallelism
alone and it takes 51 slots. Worth noting that lowering the timeout makes nothing less
graceful than today, because the container is SIGKILLed either way.

## Measurement 2 — the query plan at 100k rows

Scratch database, 100,000 rows (85k `Active`, 5k `Stopping`, 10k empty), expiries
scattered across a week, 512 due. Dropped afterwards.

| Scenario | type | key | rows | Extra |
|---|---|---|---|---|
| No composite index | `ALL` | NULL | 99006 | Using where; **filesort** |
| Composite, `status` index present | `range` | `lease_sweep` | 512 | Using where; **Using index** |
| Negative control, `lease_state` omitted | `index` | `lease_sweep` | 99060 | **filesort** |
| Composite, `status` index dropped | `range` | `lease_sweep` | 512 | Using where; Using index |

Meets the bar: `range`, the composite, hundreds of rows, no filesort, no temporary.
`ANALYZE FORMAT=JSON` gives `rows=512` against `r_rows=200` — the `LIMIT` stopping the
scan early, which is the priority-queue property the spec claims — at `r_filtered=100%`
and **0.264 ms**. The SQL is not the ceiling.

**One correction to the spec.** The README says the existing single-column `status`
index "is a strict prefix of the composite this spec adds". It is not: the composite is
`(lease_state, expires_at_ts)` and `status` is a different column. The optimizer cannot
consider it for this predicate, and the plans with and without it are identical.
**Dropping it is a cleanup, not a requirement**, and a later phase should not present it
as correctness work.

## How to verify

```bash
docker compose exec -T backend bench --site frontend execute benchpress.diagnostics.run_diagnostics
```

The `clock_skew` row must read `fail` and name the gap. A green row on a host whose
clocks differ would mean the check is not wired to anything.

```bash
docker compose exec -T backend bench --site frontend run-tests --app benchpress --module benchpress.tests.test_diagnostics
```

Note that `run-tests --app benchpress` prints **two** summaries and piping it through
`tail` returns tail's exit code, not bench's. The full suite has 13 failures
(`test_credit_sweep` 11, `test_demo_data` 1, `test_signup` 1) that are pre-existing on
`version-16` — they read live `tabBench Instance` rows. Verified on `2f7d3c3` with this
branch's source stashed.

## Nothing left behind

`tabBench Instance` is back to its pre-run count, the scratch database is dropped, the
test bench and its user are deleted, and the six tenant bench containers are untouched
and running. `enable_credits`, `max_run_hours` and `enforce_limits.stopped` all match
the preflight snapshot.


Venkateshvenki404224:
## Phase 2 — the lease clock, end to end

A deploy now buys a window. The charge lands at the moment the container reaches `Running`,
the deadline is one indexed epoch-second column, a periodic sweep claims what is due, and the
stop job confirms the deadline under a row lock before it touches Docker. The open tab
updates itself from `benchpress:lease_expired`, published to the owner's room after commit.

The hourly meter stops in this phase. Nothing in non-test code raises a burn rate, and
`reconcile_burn_rates` is no longer scheduled — it re-derived a rate from the running fleet,
which is a second billing system beside the lease.

### What landed

- `benchpress/credits/lease.py` — the only module that reasons about lease time: the catalog,
  the price, `arm`/`disarm`, the sweep, the claim, and the stop job's `FOR UPDATE` half.
- `Lease Plan` doctype, seeded 30 minutes to a week. Credits per row, not a rate, so a week is
  cheaper per hour than half an hour.
- `Bench Instance.{expires_at_ts, lease_state, stop_claimed_at, expiry_attempts}`, all
  `read_only` and `permlevel: 1`, with a composite index on `(lease_state, expires_at_ts)`.
- Per-lab configuration: `Lab.{default_lease_plan, max_lease_minutes, deploy_credits}` and
  `Instance Size.price_multiplier`.
- Frontend: `utils/lease.js` (pure), `utils/clock.js` (skew-corrected), `LeaseCountdown.vue`,
  and `socket.js` reconnecting without a cap.

### Three deviations from the spec, each deliberate

1. **The enqueue is registered before the per-claim commit, not after.** `enqueue_after_commit`
   hangs the job off the *next* commit, so enqueuing after the commit would leave it pending
   until some later, unrelated one.
2. **The single-column `status` index is kept.** Phase 1 measured the plans as byte-identical
   with and without it, and four other queries filter on `status`.
3. **The clock anchor is milliseconds (`server_now_ms`), not seconds.** A whole-second anchor
   put the browser's corrected clock up to a second behind and the countdown read ~1.85 s long.
   The deadline itself is still whole seconds.

### How to verify it

```
bench --site frontend migrate
bench --site frontend run-tests --app benchpress
cd frontend && yarn test:run
```

Live, with credits switched on the guarded way (stop `enforce_limits`, set
`max_run_hours = 0`, then `enable_credits = 1`):

- Deploy on the 30-minute plan: one `Usage` row, no burn row, `expires_at_ts` = the ledger row
  plus 1800.
- Push a deadline into the past: the bench is `Stopped` within a tick, its container is gone,
  and the open tab updates with no request between the deploy and the push.
- Start a stopped bench: the fresh deadline is written with the status, so the next sweep does
  not claim it.
- Claim a row, then move its deadline forward: the stop job no-ops and `lease_state` returns to
  `Active`.

**Not in this phase:** renew (phase 3), expiry precision better than the scheduler tick and the
dedicated stop queue (phase 4), and deleting the burn fields (phase 5).

### Measured on the live host

| Check | Result |
|---|---|
| One `Usage` row per lease, no burn row | pass — `expires_at_ts` = ledger row + 1800 exactly |
| Deadline passes → bench `Stopped`, tab updates with no request | pass |
| Charge lands at `Running`, not at deploy | pass — balance and deadline both unchanged through a 3-minute pre-`Running` window |
| **50 leases end to end, none stops early** | **pass — 0 early in 50; overrun min 31.7 s, median 33.7 s, max 62.8 s** |
| A restarted bench is not claimed on the next sweep | pass |
| Deadline moved after the claim → the stop job no-ops | pass, `lease_state` back to `Active` |
| A second tenant receives no event for the first | pass, read out of the WebSocket frames |
| Tab backgrounded past Chrome's throttling threshold | pass — countdown within a second, socket recovered |

The minimum overrun is the container stop itself, which phase 1 measured at ~31 s. Nothing
stops early, which is the acceptance bar: over-run is a budget line, stopping early is a refund.

### Two test-hygiene defects this phase found and fixed

Both are test-file-only and both damaged **live** data on a running deployment.

- `reconcile_burn_rates` is a global sweep, and four tests called it directly. On this host it
  derived a burn rate from tenant benches and charged the drift. The tests now run it through a
  helper that restores every row outside the fixture.
- `TestLease.tearDownClass` deleted its benches before its `Lease Plan` fixtures. A lock-wait
  timeout on the first delete aborted the method and left three test plans in the live catalog,
  where the lease picker offers them.

Venkateshvenki404224:
## Phase 2 — the lease clock, end to end

Two commits complete the phase and close the last gated item.

- `be24c41` `test(credits): pin the enforcement stop to the commit that decided it` — test file only, verified failing with `KeyError: 'enqueue_after_commit'`.
- `48725fd` `fix(credits): hold the enforcement stop until its decision commits`.

`credits/sweep.py:_enqueue_stop` enqueued `stop_bench` without `enqueue_after_commit=True`. That job re-reads `status`, so it could start before the decision committed and act on a row whose transaction then rolled back. `deploy_manager.py:291` already had the flag and the comment explaining it; the enforcement path did not. Both stop enqueues now carry it.

### How to verify

```
cd /home/ubuntu/benchpress_devops
docker compose exec -T backend bench --site frontend run-tests --app benchpress
cd apps/benchpress/frontend && npx vitest run
cd apps/benchpress && uvx pre-commit@4.3.0 run --all-files
```

Backend 647 + 79, 13 failures and no errors — the same 13 that predate the branch (`test_credit_sweep` 11, `test_demo_data` 1, `test_signup` 1). Frontend 174. Lint 12 hooks.

### Live results

Run against staging with credits switched on the guarded way: `enforce_limits` stopped and `max_run_hours = 0` before `enable_credits = 1`, all three restored afterwards.

| # | Item | Result |
|---|---|---|
| 1 | The pipe | PASS — one `Usage` row per lease, no burn row, deadline = charge + window |
| 2 | The clock fires | PASS — bench `Stopped` within a tick, tab updated with no request between deploy and push |
| 3 | Charge at `Running` | PASS, adapted — a cold build is unreachable (`_prepare_lab_image` throws); a 7-app site creation is the longest reachable pre-`Running` window |
| 4 | **Never early — 50 leases** | **PASS — 0 early in 50.** Overrun min 31.12 s, median 32.66 s, max 93.75 s. Windows 59–60 s, no zero deadlines |
| 5 | Resurrection control | PASS — Start writes a fresh deadline with the status; survives later sweeps |
| 6 | Renew-shaped race | PASS — claimed, stop queued, deadline moved, job no-opped, `lease_state` back to `Active` |
| 7 | Tenancy | PASS — read from the WebSocket frames, neither tenant saw the other's event |
| 8 | The hostile tab | PASS — backgrounded past Chrome's intensive throttling, correct within a second on return |
| 9 | The fleet is untouched | PASS — 6/6 running, every start time predates the run |
| 10 | Tests and lint | PASS |

The 31 s floor on item 4 is the container stop itself, which phase 1 measured at ~31 s. The sweep never claims before the deadline. Landing expiry closer than a scheduler tick is phase 4.

The hourly meter is stopped: no non-test code calls `start_burn`, `stop_burn` or `reconcile_burn_rates`, the daily hook no longer schedules the reconciler, and no `Credit Account` carries a burn rate.

Venkateshvenki404224:
## Phase 3 — Renew

A user can buy more time from the countdown, and the purchase cannot be lost to a race with the sweep trying to stop the same bench.

**`api.renew_bench(bench_name, plan, request_id)`** reads the bench row with `FOR UPDATE` — the same lock `lease.confirm_expiry` takes — and decides from that read alone. A renew either commits before the sweep's conditional claim, which then finds nothing due, or arrives after it and is refused by name. Every guard runs before the charge, so a refusal cannot debit: the `Stopping` refusal, the replay check, the lab's `max_lease_minutes` ceiling, the grace window, then the balance.

`lease.extend` pushes the deadline out from the later of now and the current deadline, so a renewal never burns time that was left. A bench stopped in its grace window has none to preserve and extends from now; renewing it starts the container it still has rather than rebuilding.

### Two defects the live race found, each fixed test-first

- **A queued expiry stop could outlive its claim.** `confirm_expiry` treated a row carrying no claim as "not this protocol's business" and waved the stop through — right for a user pressing Stop, wrong for a claimed stop sitting behind a two-hour deploy on the single `queue-long` worker. By the time it ran the bench could have been stopped, started and paid for again. `lease._enqueue_stop` now passes `from_claim=True`, and a claim-originated stop on a row that no longer carries the claim returns without touching Docker.
- **`account._locked` loaded the account with a plain read after taking the lock.** This session is `REPEATABLE-READ`, so that read answers from the snapshot the transaction opened; the stale `modified` then failed `check_if_latest`. Three parallel renewals over HTTP reproduced it — one charged, two came back `TimestampMismatchError`. One `SELECT ... FOR UPDATE` now both locks and loads.

### Idempotency

`Credit Ledger Entry.request_id` (indexed) is the key. `reference_name` could not hold it — it is a Dynamic Link and rejects a synthetic name (`LinkValidationError`, probed live) — so the ledger row is still the record, with a proper column instead of an abused link. `account.request_posted` is a locking read for the same REPEATABLE READ reason.

### Frontend

A renew dialog listing the catalog with prices, nothing selected by default, the cost on the button (`Renew 1 day · 150 credits`), one `request_id` held across retries, and no optimistic UI. Renewals arrive as a revision-ordered one-field write, so the countdown never restarts. The expiry toast is sticky. The call to action changes from Renew to Redeploy when grace ends.

### How to verify

Arm credits, deploy a bench, then:

1. **The loop** — renew for 1 day from the SPA with 5 minutes on the clock. Balance dropped 1995 → 1845 (exactly the plan price), deadline moved exactly +86400 **from the old deadline**, one new `Usage` row.
2. **The race, ten times** — sweep and renew fired at the same second, each racing for the row lock. 10/10: bench up, `lease_state` `Active`, exactly one charge, no traceback.
3. **Idempotency over HTTP** — three parallel requests, one `request_id`: one charge, one extension. Three distinct ids: three charges, +5400 total.
4. **The ceiling** — `max_lease_minutes = 60` with 50 minutes on the clock, 1-day plan refused naming the ceiling, balance unchanged to six decimal places, no ledger row.
5. **Grace** — expired and stopped, then renewed inside the window: same container id, no new deploy run, charged, started. Past the window the SPA reads "Torn down — redeploy to start again" and renew refuses.
6. **The redeploy control** — a redeploy charged a full lease and restarted the clock. The free reset is explicitly absent.
7. `run-tests --app benchpress` 671 + 79 (13 failures, all the pre-existing phase-1 baseline), `vitest run` 183, `pre-commit run --all-files` 12 hooks.

Switches, fleet and credit rows are back at their preflight values.

Still for later: expiry lands within a scheduler tick — four minutes here — and a burst drains through one worker (phase 4). The accrual engine still exists, unwritten (phase 5).

Venkateshvenki404224:
## Phase 3, iteration 2 — a deadlock on the renew path

The previous iteration's done marker was retracted by the loop's smoke check. The failing gate was `tdd_order_ok`: the phase ended on a test-only commit (`5f014bd`) that landed *after* the fix it accompanied, so no implementing commit followed the newest test commit. Every other gate passed.

Rather than manufacture a commit to satisfy it, this iteration looked for real remaining work on the renew path and found a defect.

### Two benches of one owner, renewed together, deadlocked in MariaDB

Nothing serialises two renewals on different benches — the bench rows differ — so both reach the money path at once. **5 of 5** rounds ended with one click getting `QueryDeadlockError: (1213, 'Deadlock found when trying to get lock; try restarting transaction')`, where the spec requires a sentence rather than a stack trace.

`SHOW ENGINE INNODB STATUS` gave the cycle exactly:

- `account.request_posted` reads a `request_id` that usually does **not** exist. A locking read on a missing key takes a **gap lock** on `request_id_index`.
- The ledger insert that follows needs an **insert intention** lock in that same gap.
- With the `tabCredit Account` row lock taken between the two, each transaction holds what the other needs next.

A second bug fell out of the same reading: the guard matched across the whole ledger, so **a neighbour's request id turned a renewal into a replay** — `charged: 0.0`, deadline unchanged, reported as success.

### The fix

- `account.lock(user)` takes the account row lock **before** the replay guard, so a second renewal on the same account waits there instead of reaching the gap. It deliberately does not *open* the account — doing so posted the signup grant as a side effect of a renewal that was about to be refused.
- `account.request_posted(user, request_id)` is scoped to the account, so two tenants lock different gaps.
- A `(account, request_id)` composite index (patch `index_ledger_request_id`), with `search_index` dropped from `request_id` so the single-column index cannot be chosen instead. `EXPLAIN` confirms `key: account_request_id_index`.
- `guard._account_row` reads `for_update=True`: the guard decides whether to debit, so it reads the row the debit writes.

### How to verify

```
docker compose exec -T backend bench --site frontend run-tests --app benchpress --module benchpress.tests.test_lease
```

Four new tests, landed failing in a test-only commit (`14bc689`) before the fix (`0548e0e`):

| Test | Pins |
|---|---|
| `test_the_account_is_locked_before_the_replay_guard` | the account lock precedes the ledger gap lock |
| `test_the_replay_guard_is_scoped_to_one_account` | a neighbour's request id does not swallow a purchase |
| `test_the_balance_guard_reads_the_account_under_its_lock` | no unlocked read of the balance on the renew path |
| `test_the_ledger_indexes_a_request_id_inside_its_account` | `(account, request_id)` exists |

Live on staging, all seven Verification items re-run:

| # | Item | Result |
|---|---|---|
| 1 | The loop | Deadline **+86400 from the old deadline** (from-now would have been 1787725769), balance −150 exactly, one new `Usage` row |
| 2 | Race vs. the sweep, ×10 | **10/10** — container running, `Active`, one charge, 0 stops attempted, no errors |
| 3 | Idempotency over HTTP | One id ×3 → 1 charge, **+1800 once**; three ids ×3 → 3 charges, **+5400** |
| 4 | The ceiling | Refused naming 60 minutes; balance `1,845.000000` unchanged, ledger 2 → 2 |
| 5 | Grace | Same container id, `running`, deploy runs 3 → 3. Past the window: *"This bench has been torn down. Redeploy it to start a new lease."*, balance unmoved |
| 6 | Redeploy | Charged a full 5.0 lease, one new `Usage` row, new container, deadline = `Running` + 1800 — no free reset |
| 7 | Suites | Backend **675 + 79**, 13 failures = the phase-1 baseline by name, 0 errors. vitest **183**. pre-commit **12/12** |

**The deadlock race, re-run after the fix: 20/20 clean** — 10 rounds with both renewals affordable (2 charges, 0 deadlocks) and 10 with only one affordable (1 charge, 1 honest shortfall naming the amount, 0 deadlocks).

Fleet, switches and credit rows are byte-equal to preflight: 6/6 containers, `0` / `8` / `0`, zero credit accounts, zero ledger rows.

Venkateshvenki404224:
## Phase 4 — Precision and drain

Nothing a user sees changes. The lease behaves as phase 3 left it; it ends closer to when it
said it would.

### What this adds

**A lease warden** (`benchpress/credits/warden.py`, plus a `lease-warden` service in
`benchpress_devops`). A long-lived loop: read the next deadline, sleep until it or at most 15
seconds, claim, hand the stop over. The scheduler cannot do this — `DEFAULT_SCHEDULER_TICK` is
four minutes here, so even a `*/1` cron entry fires every four. It runs on every node and the
conditional claim arbitrates, so there is no election and no failover gap.

**The Frappe cron stays**, folded into the existing `*/5` entry as the net under the warden
while it restarts. Two mechanisms, one claim protocol.

**A dedicated stop queue** (`queue-stops`, `STOP_WORKERS` workers on the `stops` queue). Stops
used to share `long` with `deploy_bench`, which holds the single worker for up to two hours.

**A configurable stop grace period**, defaulting to 5 s. Phase 1 measured that PID 1 in these
images installs no SIGTERM handler, so every stop waits the grace period out in full and then
SIGKILLs. The old 30 s bought nothing and was six times the drain rate.

**`Bench Instance.node`**, and `lease.assert_local` in front of every Docker-touching stop.
`stop_container` swallows `NotFound` as success, so a stop routed to the wrong daemon would
mark the row `Stopped` while the container kept running elsewhere. Empty on every row today,
so routing changes nothing yet.

**A backlog gauge, a reclaim and a park** (`benchpress/credits/drain.py`). If leases fall due
faster than `cap × cadence`, every one of them overruns forever and no individual sweep fails —
the gauge is the only alarm there can be, and it logs on the second consecutive overflow, not
on every sweep. A worker killed mid-batch strands its rows in `Stopping`; the reclaim re-queues
them and parks the ones no retry will fix.

**Instrumentation.** `stop_started_at`, `container_stopped_at` and `expiry_lateness` per stop.
`drain.stop_slo()` reads the histogram back.

### Measured on staging

| | before | after |
|---|---|---|
| Deadline to container stopped, 10 leases (min / median / max) | 185 / 191 / 196 s | **8 / 14 / 20 s** |
| 100 leases expiring in the same second | 50.9 min (phase 1) | **164 s** |

The two levers are separable: the 5 s grace is 6x, the 4-worker pool is ~3x.

Crash recovery: `docker kill` on the stop worker with 50 rows in `Stopping` left nothing
stranded and nothing parked — clear in 6 minutes at a 60 s reclaim interval, attempts peaking
at 3 against a limit of 5.

Batch cap with 100 due and `lease_sweep_batch = 5`: exactly 5 claimed per sweep, overflow
reported at 95 / 90 / 85, and one Error Log row rather than one per sweep.

Never early: **200 leases across 4 rounds, 0 early**, minimum overrun 5.2 s.

Wrong node: a stop for `node = "nowhere"` raises by name and leaves the row `Running` and the
container running.

### How to verify

```bash
docker compose ps lease-warden queue-stops
docker compose exec -T backend bench --site frontend execute benchpress.credits.drain.stop_slo
docker compose exec -T backend bench --site frontend execute benchpress.credits.drain.sweep_expired_leases
```

The sweep returns `{"due", "cap", "overflow", "claimed", "reclaimed"}` and does nothing at all
when `enable_credits` is off.

Rollback needs no code revert: stop the warden and the cron alone is phase 3's behaviour;
remove the `workers` key and stops fall back to `queue-long`.

### Tests

`bench --site frontend run-tests --app benchpress` — **719 + 79**, 0 errors, 13 failures, all
the pre-existing baseline (`test_credit_sweep` 11, `test_demo_data` 1, `test_signup` 1: they
assert against the live fleet, which exceeds `max_run_hours`).
`npx vitest run` — 183. `uvx pre-commit@4.3.0 run --all-files` — 12 hooks, all passed.

The compose half is in `benchpress_devops@edd4be5`.

Venkateshvenki404224:
## Phase 5 — delete the meter

The accrual engine is out of the tree. `start_burn`, `stop_burn`, `settle`, `correct_burn_rate`,
`accrued`, `credits/reconcile.py`, `credits/passes.py`, the `Always On Pass` doctype, the
`max_run_hours` TTL and every frontend burn label are gone. `available()` is the stored balance,
so nothing on a read path moves with the clock.

`guard` now proves the price of one lease instead of an hour at a rate no code computes, and the
size picker quotes the same number — what the deploy is about to spend.

Two patches carry the schema:

- `retire_always_on_passes` (pre_model_sync) converts every live pass into a lease of at least its
  remaining days, then deletes the doctype and drops its table. `delete_doc` does not drop a
  DocType's table, so the patch does it by hand.
- `drop_burn_fields` (post_model_sync) drops `burn_rate`, `burn_since`, `credit_burn_rate`,
  `credit_burn_started` and `ttl_warned_at`. Both return what they changed, so a second run says
  nothing.

**One deliberate exception.** `max_run_hours` and `always_on_monthly_inr` leave
`credit_settings.json`, `NON_NEGATIVE_FIELDS` and every code path — the DocField goes on the next
sync — but their stored `tabSingles` values are left where `bench migrate` leaves them. Frappe
never removes a Single's value on its own, an orphan value nothing declares is inert, and the
loop's own safety gate reads `max_run_hours` to prove no verification run left the switches
flipped.

### How to verify

```
docker compose exec -T backend bench --site frontend run-tests --module benchpress.tests.test_meter_gone
docker compose exec -T backend bench --site frontend run-tests --app benchpress
cd frontend && yarn test:run
uvx pre-commit@4.3.0 run --all-files
```

`test_meter_gone` scans every `.py` and `.json` in the app for the retired names, asserts the
columns and the table are gone, asserts `available(row) == row.balance` even for a row still
carrying burn fields, and asserts every account's ledger sums to its balance.

### Measured on staging

| Check | Result |
|---|---|
| Balances across `bench migrate` | `Administrator 30.0`, `lc-p5-tenant 400.0` — byte-identical before and after |
| Pass conversion | a pass with 12 days left became a lease with **12.400 days** left, `lease_state=Active` |
| Schema | `tabAlways On Pass` dropped; no burn column on `tabCredit Account` or `tabBench Instance` |
| `bench migrate` twice | both clean; the second run prints no patch line |
| Ledger | every account reconciles, drift `[]`; the credit reconciliation report runs, 0 unsettled |
| Deploy → expire, credits on | lease charged 5 credits, deadline 30 min, stopped **19 s** after it — never early |
| Backend | 702 + 79, 2 failures — the `test_demo_data` / `test_signup` baseline, unchanged |
| Frontend | 184 passed |
| Lint | 12 hooks, all pass |

The six tenant containers were untouched throughout, and `enable_credits`, `max_run_hours` and the
`enforce_limits` job are back where preflight found them.

Venkateshvenki404224:
## Live UI audit on staging

Driven through a real browser against `staging.benchpress.cloud` with credits switched on. Every screenshot below is a capture of the running system, not a mockup.

### What works

**The countdown renders in both places the plan specified.**

Labs list — the deadline sits under the instance name, and only a bench that holds a lease shows one:

Labs list countdown

Lab detail, Container card — beside the existing "Started N ago" line, with Renew next to it:

Container card countdown

**The renew picker** lists the catalog with prices, and states the extend-from-deadline rule in words: *"Time is added to the deadline you already have."*

Renew plan picker

**The refusal path names the shortfall and charges nothing.** Selecting a 700-credit week against a 40-credit balance:

Insufficient credits

Balance stayed at 40. Negative control passes.

**A real renew extends from the old deadline, not from now.** 28:25 → 56:53 on a 30-minute renew, balance 40 → 35, no page reload:

Renew extended

**The Credits page carries lease language,** and the ledger row reads `bench-client-frappe-16 — 30 minutes lease`:

Credits ledger

Also verified in the database rather than the UI:

- A real redeploy charged 5 credits and armed a 30-minute lease on reaching `Running`.
- `on_bench_running` is idempotent (`lease_state == ACTIVE` short-circuits), so a repeated transition cannot double-charge.
- The renew debit carries a `request_id` (`6eadc741-…`); the deploy debit does not need one because the charge is keyed on the status transition.
- `requires_credits` only refuses — it never debits — so a double-click on Deploy cannot spend twice.

### Gaps worth a follow-up

1. **Deploy and redeploy never name their price.** `LabHeader.vue` `CONFIRMATIONS.deploy` warns only about data being destroyed. Renew shows `Renew 30 minutes · 5 credits`; redeploy spends the same 5 credits and says nothing. This is the one a user notices first — it reads as "it didn't ask me".
2. **Unaffordable plans are selectable.** The picker enables `Renew 1 week · 700 credits` at a 40-credit balance and only fails after the click. The client already knows the balance, so those rows could be disabled or annotated with the shortfall.
3. **Overview has no lease surface at all.** It reports "6 running" with nothing about what expires next — the one page a user lands on first.

None of the three is a correctness bug; all three are the same omission, which is that spend and time-remaining are not surfaced where the decision is made.

Venkateshvenki404224:
## A paid renewal could be silently reverted — found live, fixed

A user renewed for 8 hours, was charged 60 credits, and the countdown never moved. Money taken, time not delivered.

### Root cause

The deploy log dates it exactly:

| Time | Event |
|---|---|
| 16:13:54 | redeploy job starts — loads the `Bench Instance` document |
| 16:16:10 | **renew commits** a new `expires_at_ts` and charges 60 credits |
| 16:16:36 | deploy finishes (162.0s) and calls `bench.save()` |

`lease._write` updates the row with `update_modified=False` and deliberately bypasses `save()`. But the deploy job held a document loaded **two minutes earlier**, and `save()` writes every field from that copy — including the deadline the renewal had just moved. The bench row's `modified` is 16:16:36, 26 seconds after the charge.

Nothing about this is specific to renew. Any job holding a `Bench Instance` for a while could revert any lease write, and `deploy_manager` has ten `bench.save()` calls.

### Fix

`BenchInstance.before_save` re-reads the four lease-owned columns (`expires_at_ts`, `lease_state`, `stop_claimed_at`, `expiry_attempts`) from the row, so `lease._write` is their only writer and no stale document can carry them. Verified first that nothing else assigns them — every other reference in the app is a comparison.

Landed test-first: the regression test loads a document, renews through another path, saves the stale copy, and asserts the deadline survived. It fails on the old code with `1787657013 != 1788261813`.

### Verified live, not just in tests

- **5-minute lease armed on reaching Running:** deadline `1787655750` against an expected `1787655748` — 2s, the time the start took. The clock arms at `Running`, never at the call.
- **Expiry:** `Active → Stopping → Stopped`, container gone **8 seconds** past the deadline. That is the `lease-warden` loop; the 4-minute scheduler tick would have given away 13% of a 30-minute lease.
- **Paying just before the deadline, 6 of 6 runs:** renew won, charged exactly 1.00, deadline moved, bench stayed up. One run extended by 306s rather than 300 because the deadline had already passed — correctly extending from now rather than from a stale deadline.
- **Paying just after the claim:** refused by name — *"This bench is already stopping."* Balance unchanged, zero ledger rows for that request id.

### Also in this push

**A Playwright suite for the countdown** (`e2e/lease-countdown.spec.ts`, 5/5 against staging): that it renders, that it ticks rather than holding the first paint, that a bench with no lease shows nothing, and that client zero says "Lease ended" without claiming the bench stopped. Two of them initially passed for the wrong reason — a countdown on another lab's row answered the assertion — so they now filter the table to the lab they create.

**A test that could not fail.** `test_credit_guard` hardcodes `LEASE_COST = 5.0` but took the actual price from `default_lease_plan`, which is admin-editable. Pointing the site at a cheaper plan turned every shortfall assertion into a pass — which is how it was found: setting a 1-credit default made `test_a_balance_below_the_lease_price_is_a_shortfall` sail past the guard and into a 404 from Docker. The module now owns a plan priced at `LEASE_COST`, and the suite passes with the site default deliberately set to the cheap plan.

### Suite state

703 tests. Two failures remain and neither is from this change:

- `test_a_seeded_run_lasts_minutes_not_weeks` — asserts against live `Deploy Log` rows and trips on one created 2026-08-20 that spans 3630s against a 3600s limit.
- `test_a_retirement_notice_goes_out_once_per_entry` — fails identically at the pre-fix commit; confirmed by reverting.

Both read live site state rather than their own fixtures, the same class of fragility as the guard test above.

PR #161, opened 2026-08-25
Implemented phase three

Venkateshvenki404224:
Phase 1 of the **atomic-admission** spec: admission becomes a write that can fail.

## What changed

`cap_concurrent_instances` counted the caller's `Running` benches and compared. Two requests that arrived together read the same count and both passed, and `create_bench` writes its row as `Draft` for the two minutes a deploy takes — so an in-flight deploy was invisible to the thing counting deploys.

A slot is now a row. `Bench Admission` autonames on the bench, which puts the claim on a primary key, and both the read and the write happen under `SELECT ... FOR UPDATE` on the caller's `Credit Account` — the one row every contender for that caller has to take.

- **`Credit Settings.max_concurrent_uncredited`**, default `0`. The cap that applies when credits are off. `0` means unlimited, so an install that has not opted in behaves exactly as it did.
- **The claim runs whatever the credits switch says**; only the balance check stays behind it. `requires_credits` is now `requires_admission`.
- **Every path that ends a bench gives the slot back**: deploy failure, `stop_bench`, `teardown_bench`, `on_trash`. `_redeploy_bench` passes `release_admission=False` so a redeploy keeps its own slot across the teardown.
- **`credits.admission_repair.reconcile_admissions`** on the `*/5` cron: expires stranded claims, adopts live instances holding none, heals the counter from the rows.
- `account._locked` is now `account.locked`, and the signup grant is idempotent by ledger reference, because admission opens accounts on sites with credits switched off.

## How to verify

Baseline, on the parent commit (behaviour files reverted, harness kept):

```
admitted 12 / attempted 12 · cap 1 · refused 0 (cap) 0 (other) · 0 errors
account: active_instances=0 rows=0
FAIL: expected 1 admitted, got 12          exit 1
```

Same command on this branch:

```
python3 scripts/admission_drill.py --mode cap --workers 12 --cap 1 --i-know-this-is benchpress.cloud

admitted 1 / attempted 12 · cap 1 · refused 11 (cap) 0 (other) · 0 errors
account: active_instances=1 rows=1         exit 0
```

Also measured against the live staging host:

- cap `2`, twelve workers -> `admitted 2`; cap `0` -> `admitted 12` with twelve `Bench Admission` rows, so `0` still claims.
- `SELECT active_instances FROM tabCredit Account` and `SELECT COUNT(*) FROM tabBench Admission` agree at every step, and both read `0` after `--cleanup`.
- A real deploy held one claim, with `claimed_at` unchanged, from the request through `Deploying` to `Running` and across two `enqueue_redeploy` calls.
- `queue-long` killed mid-deploy: no `except` block ran, and one `*/5` tick of the repair pass moved the bench to `Error`, dropped its row and returned the counter to `0`.

`bench run-tests --app benchpress`, `vitest run` and `uvx pre-commit@4.3.0 run --all-files` all pass. Two suite failures predate this branch and fail identically on `version-16`: `test_demo_data.test_a_seeded_run_lasts_minutes_not_weeks` and `test_signup.test_a_retirement_notice_goes_out_once_per_entry`.

## Two deviations from the phase spec, both deliberate

- The repair module is `credits/admission_repair.py`, not `credits/reconcile.py`. `test_meter_gone` asserts the name `credits.reconcile` never comes back — it was the burn meter's repair pass.
- The spec's worker-death check expects one tick to error an abandoned deploy, which contradicts its own `STALE_DEPLOY_HOURS = 3` rule. Verified with `claimed_at` back-dated past that window; the repair still came from the reconciler and from no `except` block.

Still open by design: the runway is a check rather than a hold, and `api.bench_action` is ungated (phase 2); two benches can still win the same `site_name` (phase 3).

### DIFF
--- a/benchpress/api.py
+++ b/benchpress/api.py
@@ -24,11 +24,10 @@ from benchpress.credits import account, config, lease, metering, payments
 from benchpress.credits.guard import (
 	build_charge,
 	cap_builds_per_day,
-	cap_concurrent_instances,
 	cap_devices,
 	payload_lease_cost,
 	require_balance,
-	requires_credits,
+	requires_admission,
 )
 from benchpress.permissions import (
 	get_bench_owner_filter,
@@ -83,7 +82,7 @@ def create_lab_from_template(template: str, lab_id: str | None = None, title: st
 
 
 @frappe.whitelist()
-@requires_credits(cost=build_charge, caps=(cap_builds_per_day,))
+@requires_admission(cost=build_charge, caps=(cap_builds_per_day,))
 def build_lab_image(lab_name: str) -> dict:
 	require_admin()
 	frappe.enqueue(
@@ -164,7 +163,7 @@ def _counts_by_bench(doctype: str, column: str, bench_names: list[str]) -> dict[
 
 
 @frappe.whitelist()
-@requires_credits(cost=payload_lease_cost, caps=(cap_concurrent_instances,))
+@requires_admission(cost=payload_lease_cost)
 def create_bench(data: str) -> dict:
 	require_app_user()
 	from benchpress.benchpress.doctype.bench_instance import get_instance_id
@@ -380,7 +379,7 @@ def get_deploy_history() -> dict:
 
 
 @frappe.whitelist()
-@requires_credits(caps=(cap_devices,))
+@requires_admission(caps=(cap_devices,))
 def add_device(device_name: str, device_type: str, public_key: str | None = None) -> dict:
 	require_app_user()
 	from benchpress.vpn_adapter import register_device
--- a/benchpress/api.py
+++ b/benchpress/api.py
@@ -179,40 +179,20 @@ def create_bench(data: str) -> dict:
 	lab = frappe.get_cached_doc("Lab", lab_name)
 	requested_site_name = data.get("site_name") or data.get("site")
 
+	site_name = resolve_site_name(requested_site_name)
 	instance_id = get_instance_id(frappe.session.user, lab_name)
 	if frappe.db.exists("Bench Instance", instance_id):
-		doc = frappe.get_doc("Bench Instance", instance_id)
-		site_name = resolve_site_name(requested_site_name, exclude_bench=doc.name)
-		if site_name and site_name != doc.site_name:
-			_assert_site_name_changeable(doc)
-			doc.site_name = site_name
-		doc.status = "Deploying"
-		doc.save()
+		doc = _redeploy_instance(instance_id, site_name)
 	else:
-		doc = frappe.get_doc(
-			{
-				"doctype": "Bench Instance",
-				"bench_name": data.get("bench_name"),
-				"lab": lab_name,
-				"frappe_version": lab.frappe_version,
-				"domain": data.get("domain"),
-				"site_name": resolve_site_name(requested_site_name),
-				"status": "Draft",
-			}
-		)
+		try:
+			doc = _new_instance(lab, data, site_name)
+		except frappe.DuplicateEntryError:
+			# A call for the same (user, lab) that arrived first already named this bench.
+			# `get_instance_id` is deterministic, so the branch was always meant to be idempotent.
+			frappe.clear_last_message()
+			doc = _redeploy_instance(instance_id, site_name)
 
-		for app in lab.apps:
-			doc.append(
-				"apps",
-				{
-					"app_name": app.app_name,
-					"app_label": app.app_label,
-					"git_url": app.git_url,
-					"branch": app.branch,
-				},
-			)
-
-		doc.insert()
+	_claim_site_name(doc)
 
 	frappe.enqueue(
 		"benchpress.deploy_manager.deploy_bench",
@@ -225,6 +205,77 @@ def create_bench(data: str) -> dict:
 	)
 
 	return {"name": doc.name, "status": "Deploying"}
+
+
+def _redeploy_instance(instance_id: str, site_name: str | None):
+	doc = frappe.get_doc("Bench Instance", instance_id)
+	if site_name and site_name != doc.site_name:
+		_assert_site_name_changeable(doc)
+		doc.site_name = site_name
+	doc.status = "Deploying"
+	doc.save()
+	return doc
+
+
+def _new_instance(lab, data: dict, site_name: str | None):
+	doc = frappe.get_doc(
+		{
+			"doctype": "Bench Instance",
+			"bench_name": data.get("bench_name"),
+			"lab": lab.name,
+			"frappe_version": lab.frappe_version,
+			"domain": data.get("domain"),
+			"site_name": site_name,
+			"status": "Draft",
+		}
+	)
+	for app in lab.apps:
+		doc.append(
+			"apps",
+			{
+				"app_name": app.app_name,
+				"app_label": app.app_label,
+				"git_url": app.git_url,
+				"branch": app.branch,
+			},
+		)
+	return doc.insert()
+
+
+def _claim_site_name(bench) -> None:
+	"""Claim the site name by insert. The primary key is the check, and nothing else is.
+
+	The name is held for exactly as long as the row: `teardown_bench` keeps it so a redeploy
+	still owns its own name, and only `_delete_bench` frees it - which is when the database
+	behind it actually stops existing.
+	"""
+	try:
+		site = frappe.get_doc(
+			{
+				"doctype": "Bench Site",
+				"site_name": bench.site_name,
+				"bench": bench.name,
+				"status": "Creating",
+			}
+		).insert(ignore_permissions=True)
+	except frappe.DuplicateEntryError:
+		_refuse_taken_site_name(bench)
+		return
+	# After the insert, which stamps the session user over any owner handed to it. `Bench Site`
+	# is read `if_owner`, so an admin deploying somebody else's instance would otherwise take the
+	# row over and empty that tenant's Sites tab.
+	frappe.db.set_value("Bench Site", site.name, "owner", bench.owner, update_modified=False)
+
+
+def _refuse_taken_site_name(bench) -> None:
+	# First, and on both paths: the framework msgprints "Bench Site X already exists" before it
+	# raises, so a redeploy would carry that internal sentence on an otherwise successful reply.
+	frappe.clear_last_message()
+	# A locking read, because this session is REPEATABLE READ: a plain one answers from a
+	# snapshot taken before the winner committed, and would refuse this caller their own name.
+	if frappe.db.get_value("Bench Site", bench.site_name, "bench", for_update=True) == bench.name:
+		return
+	frappe.throw(_("Site name '{0}' is already in use. Choose a different name.").format(bench.site_name))
 
 
 def _assert_site_name_changeable(doc) -> None:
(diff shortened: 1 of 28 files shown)


Venkateshvenki404224:
Follow-up commit `f98c878` — the drill harness could not hand out its own credentials.

`drill.setup` opens the drill user, mints twelve labs and sets the concurrency cap in one call, so anything that wanted only the token got the other two as side effects. There was no way to ask for the token on its own, and `_ensure_user` also regenerated the API secret on every call, so a token read before a run no longer authenticated during it.

`drill.ensure_drill_user()` now does the user half alone and returns `key:secret`; the secret is kept once minted.

Verify:

```bash
docker compose exec -T backend bench --site frontend execute benchpress.credits.drill.ensure_drill_user
# prints key:secret, and prints the same pair on a second call
```

Re-measured against the live endpoint after the change: `admitted 1 / attempted 12 · cap 1 · refused 11 (cap) 0 (other) · 0 errors`, exit 0, `active_instances=1 rows=1`, both back to `0` after `--cleanup`.

Test suite 25 tests in `test_admission.py`, including one that pins the token shape the harness reads. Frontend 184 pass. `pre-commit` clean. The two failures in the full backend run (`test_demo_data.test_a_seeded_run_lasts_minutes_not_weeks`, `test_signup.test_a_retirement_notice_goes_out_once_per_entry`) fail identically on `version-16` and are untouched by this branch.

Venkateshvenki404224:
## Phase 2 — a hold, not a check

One hour of runway stops being a number the gate reads and becomes a number the gate holds.

**What it adds**

- `Credit Account.reserved_credits` and `Bench Admission.held_credits` — the hold, denormalised on the account and stored per bench.
- `account.spendable()` — the balance less what is already held. `available()` is unchanged and stays what the sweep asks: a hold must never read as running out.
- `admission.claim(user, bench, limit, cost)` decides the cap **and** the credits under one `SELECT … FOR UPDATE` on the caller's `Credit Account`. `guard._require_balance` is gone from that path; `require_balance` survives for the renew endpoint and now also reads `spendable`.
- `admission.release_hold()` ends the hold beside the charge in `metering.on_bench_running`, under the lock that charge just took. `release()` gives the hold back with the slot on every exit — failure, stop, teardown, delete, and the repair pass.
- `api.bench_action` start and restart are behind the gate. It is the path the SPA uses and carried no limit at all. `stop` stays open.
- `admission_repair` heals `reserved_credits` from the `held_credits` it aggregates.
- `scripts/admission_drill.py --mode credits`.

A hold is not a charge: no balance moves and no `Credit Ledger Entry` is written for it.

**How to verify it**

```bash
cd apps/benchpress
AUTH=$(docker compose exec -T backend bench --site frontend execute benchpress.credits.drill.ensure_drill_user | grep -oE '[A-Za-z0-9]{8,}:[A-Za-z0-9]{8,}')
BENCHPRESS_DRILL_AUTH=$AUTH python3 scripts/admission_drill.py --mode credits --workers 12 --cleanup --i-know-this-is benchpress.cloud
BENCHPRESS_DRILL_AUTH=$AUTH python3 scripts/admission_drill.py --mode cap --workers 12 --cap 1 --cleanup --i-know-this-is benchpress.cloud
```

Measured on staging:

| Step | Result |
|---|---|
| 12 parallel `create_bench`, 3 credits, labs at 1.0 | `admitted 3 / attempted 12 · refused 0 (cap) 9 (credits) · 0 errors` |
| The hold in the table while the three deploy | `reserved_credits 3.0`, three rows at `1.0`, `balance` still `3.0`, ledger holds only the signup Grant |
| `summary()` | `{"balance": 3.0, "reserved": 3.0, ...}` |
| Conversion (`--allow-deploys`, real image, bench reaches `Running`) | `reserved_credits 1.0 → 0.0`, `held_credits → 0.0`, `balance 3.0 → 2.0`, exactly one `Usage -1.0` row, the slot survives |
| Failed deploy (image the lab never built) | three benches `Error`, rows gone, `reserved_credits → 0.0`, `balance` untouched, no ledger row for the hold |
| `bench_action` `start` / `restart` at the cap | refused with the cap sentence; with the decorator removed the same call reaches the endpoint body |
| `bench_action` `stop` | not gated — reaches the endpoint's own message |
| cap drill, unchanged from phase 1 | `admitted 1 / attempted 12 · refused 11 (cap)`, `reserved=1.0 held=1.0` |
| Fleet | six pre-existing benches untouched; `enable_credits`, the sweep job and every cap back as found |

`run-tests --app benchpress`: 744 tests, the same 2 failures `version-16` has (`test_demo_data`, `test_signup`). `vitest run`: 184 pass. `pre-commit --all-files`: pass.

**Deviations from the phase spec** — it was written against the burn-rate meter, which leases replaced.

- No `rate_per_hour` on the claim. There is no rate any more; a lease is one charge, so `held_credits` is the whole hold.
- `convert_hold` is `release_hold`: the hold does not become a burn, it ends when `account.charge` debits the lease.
- The repair pass is `admission_repair`, not `credits/reconcile` — `test_meter_gone` asserts that name never comes back.
- `shortfall_message` moved to `account`, not `guard`: `guard` imports `admission`, so the sentence had to sit where both can reach it without a cycle.
- The sweep, the TTL and `enable_credits` were **not** flipped. Credits are already on, `max_run_hours` no longer exists, and no tenant container was running — flipping them would have failed the loop's own switch check.

**Known gaps, named rather than hidden**

- A bench that already holds a slot (a redeploy or restart of something live) is priced but not held. Two such redeploys by one caller can still overspend — the same shape as `cap_builds_per_day`, which still counts ledger rows written after a 10–40 minute build.
- `summary()` returns `reserved`; the SPA meter does not render it yet.
- Two benches can still win the same `site_name`. That is phase 3.

Venkateshvenki404224:
## Phase 3 — the site name becomes a key

`site_name` had no unique constraint. `docker_manager._assert_site_name_available` checked it in the request; `_record_primary_site` wrote the `Bench Site` row from a worker about two minutes later. The name keys a MariaDB database through `"_" + sha1(site_name)[:16]`, and `mariadb_manager` creates it with `CREATE OR REPLACE DATABASE` — so two benches that won the same name shared one database and the second dropped the first tenant's data.

**Measured on staging, twelve simultaneous `create_bench` calls asking for one name:**

| | admitted | refused | instances claiming the name |
|---|---|---|---|
| before | **12** / 12 | 0 | **12** |
| after | **1** / 12 | 11 (site name taken) | **1** |

### What changed

- **`Bench Site` autonames on `field:site_name`.** A primary key, not `unique: 1`, so a collision raises `frappe.DuplicateEntryError` — the exception `credits.admission._insert` already handles.
- **`api._claim_site_name` claims the name in the request**, right after the `Bench Instance` insert and before `frappe.enqueue`. `enqueue_after_commit=True` means the refusal rolls the instance back and queues nothing: the eleven losers left no `Bench Instance` row and no `Error Log` entry.
- **`docker_manager._assert_site_name_available` is gone.** `resolve_site_name` keeps normalisation and validation only. The old check ran in the wrong process and excluded `Inactive` rows, so it handed out the name of a stopped bench whose database was still on disk.
- **`patches.name_bench_sites_by_site_name`** moves existing rows onto the key. A duplicate loser is suffixed and set `Inactive`, **never deleted** — an `Inactive` row can still own a live database, and `api._drop_bench_site_databases` reads `site_name` to find what to drop. Staging had 7 rows and no duplicates, so the dedupe branch did not fire there; it is covered by a planted collision in the tests.
- **`_record_primary_site` activates the claimed row** and raises if the row belongs to another bench, rather than writing over it.
- **`admission_repair` retires a claim left `Creating` past `STALE_DEPLOY_HOURS`**, turning `patches.retire_orphaned_creating_sites` into a standing rule now that a request can create that state again.
- **`scripts/admission_drill.py --mode site-name`** — N distinct labs, one site name.

### The deliberate tightening

**A name stays claimed for as long as its row exists.** `teardown_bench` keeps the row on purpose, and a redeploy of the *same* bench still works because the row is its own — verified twice over HTTP, both 200, the row keeping its name and its `creation`. Another bench can only take the name after `api._delete_bench` removes it. That is stricter than before, and it is the correct strictness: it matches when the database actually stops existing.

### Two bugs the tests found, both fixed here

- **`insert()` stamps the session user over any `owner` handed to it.** `Bench Site` is read `if_owner`, so an admin deploying somebody else's instance would have taken the row over and emptied that tenant's Sites tab. The owner is now pinned after the insert.
- **The redeploy path leaked the framework's message.** Frappe `msgprint`s "Bench Site X already exists" *before* it raises, so a successful redeploy carried that internal sentence in `_server_messages`. `frappe.clear_last_message()` now runs on both paths, not only before the throw.

### How to verify

```bash
# the key exists and no row was lost
SELECT COUNT(*), SUM(name <> site_name) FROM `tabBench Site`;   -- 7, 0

# the property, through the shipped endpoint
python3 scripts/admission_drill.py --mode site-name --workers 12 --cleanup --i-know-this-is <base_domain>
# admitted 1 / attempted 12 · refused 11 (site name) · 0 errors
# site '<name>': 1 rows · 1 instances claim it

# the refusal is one sentence, not two
curl -s -X POST .../api/method/benchpress.api.create_bench -d '{"data":"{\"lab\":\"<lab>\",\"site_name\":\"<taken>\"}"}'
# HTTP 417, one _server_messages entry: "Site name '...' is already in use."
```

Ran on staging with `--allow-deploys`: exactly one container, and exactly one new schema `_cdea2e591883efee` on the tenant MariaDB. Cleanup dropped it; the seven pre-existing `Bench Site` rows and six tenant instances were untouched throughout.

Backend suite 759 tests, frontend 184, `pre-commit --all-files` clean. Two failures are known on `version-16` (`test_demo_data.test_a_seeded_run_lasts_minutes_not_weeks`, `test_signup.test_a_retirement_notice_goes_out_once_per_entry`). One `test_lease.TestLease` test errors in a full run and passes when the module runs alone; the same class errors the same way without this branch's changes — the suite leaks `BenchPress Settings.enable_credits` and Administrator's balance across modules. Not introduced here, and worth its own issue.

### Still open, deliberately

`cap_builds_per_day`, `cap_devices` and `cap_sites_per_instance` still count before they act; `Always On Pass` and the default `Instance Size` can still be written twice; the reconciler heals within five minutes rather than immediately. A redeploy or restart of a bench that already holds a slot is priced, not held.

Venkateshvenki404224:
### Follow-up: `2bdb42d` — the name is freed when the instance is deleted

Found while checking the staging row count after a full suite run. Two `Bench Site` rows were left behind, one owned by `lease-owner@example.com` and one by `authz-user-a@example.com`, both pointing at benches that no longer existed and neither backed by a database.

The cause is this phase's own doing. Before it, only a *deployed* bench had a `Bench Site` row; now every instance owns one from birth. So any delete that skipped `api._delete_bench` — `force=True` in a test, `drill._delete_bench` — stranded a row holding a name for a bench that was gone.

Such a row can never drop its own database: `_drop_bench_site_databases` is reached only from `_delete_bench`, and that needs the instance. Keeping it costs a name and buys nothing. `BenchInstance.on_trash` now removes it beside the concurrency slot, where every delete path passes, and `_delete_bench` drops the databases before that while the rows are still there.

The fix reclaimed both stranded rows on its own, through the same teardown that had been leaking them. Staging is back to exactly the 7 `Bench Site` rows it had before this phase.

Backend suite re-run: **760 tests, 2 failures**, both known on `version-16` (`test_demo_data.test_a_seeded_run_lasts_minutes_not_weeks`, `test_signup.test_a_retirement_notice_goes_out_once_per_entry`). The `test_lease` error mentioned above is gone with this change.

PR #164, opened 2026-08-25
Implemented phase three

Venkateshvenki404224:
## Phase 1 — a bench lands on a named bridge

Every bench joined one Docker network, `benchpress` on `172.30.0.0/24` — 253
addresses shared with Traefik, MariaDB and Redis, behind a bridge device whose
name changes whenever the network is recreated. This is the first slice of the
`density-kernel-limits` spec.

### What changed

- `docker_manager` gains the family arithmetic: `bench_network_spec(i)` →
  network `benchpress-<i>`, device `bpbr<i>`, subnet `10.20.<i*16>.0/20`,
  gateway `.1`. `ensure_bench_network` creates a bridge and attaches the three
  infrastructure containers; `attach_infrastructure` tolerates a missing
  container so a dev checkout with no Traefik still deploys.
- `Bench Instance.bridge_network` — `Data`, `permlevel: 1`, read-only, defaulted
  in `validate` and refused outside `Draft`, copying the `runtime` pattern.
- The container create, both IP read-backs and the WireGuard endpoint take the
  network they act on instead of naming `benchpress`.
- `backfill_bench_bridge_network` stamps existing rows with `benchpress`, the
  network their containers are actually on. Nothing migrates.
- `entry.py` creates `benchpress-0`, and Traefik declares it as
  `benchpress-bench-0` (in `benchpress_devops`, branch of the same name).

`INFRASTRUCTURE_CONTAINERS` is a closed tuple rather than a setting: the control
plane's own database and caches must never sit on a network tenants share.

### How to verify

```bash
docker network inspect benchpress-0 --format '{{range .Containers}}{{.Name}} {{.IPv4Address}}{{"\n"}}{{end}}'
ip -br link | grep bpbr0
ip -br route | grep 10.20.0.0/20
curl -sI https://<instance_id>.<base_domain>/api/method/ping    # 200
docker exec <id> getent hosts benchpress-mariadb benchpress-redis
docker exec <id> getent hosts benchpress_db                     # must resolve nothing
docker exec <id> grep Endpoint /etc/wireguard/wg0.conf          # 10.20.0.1
```

Verified live on staging: a bench deployed onto `benchpress-0` at `10.20.0.5`
behind `bpbr0`, served `{"message":"pong"}` over HTTPS through Traefik, resolved
both infrastructure containers, did not resolve `benchpress_db`, and held a
WireGuard handshake against `10.20.0.1`. `benchpress-0` carries exactly four
endpoints. A bench pinned to the legacy `benchpress` network deployed and served
alongside it, and all six pre-existing rows read `benchpress`.

### Known gaps (phases 2 and 3)

Index 0 is the only bridge anything chooses, nothing reports how full it is,
nothing re-attaches infrastructure after a `compose up` recreates Traefik, and
no kernel ceiling has moved.

### DIFF
--- a/.claude/skills/issue-to-phases/SKILL.md
+++ b/.claude/skills/issue-to-phases/SKILL.md
@@ -100,15 +100,21 @@ else the repo default.
 3. Implement from the phase spec. Follow the repo conventions, run the lint hook.
 4. Commit, `git push -u origin <branch>`
 5. `gh pr create --base <base> --head <branch> --title "Implemented phase one" --body "<what it does + how to verify>"`
+6. `gh pr checks <branch> --watch` — fix whatever is red before calling the phase done.
 
 **Phase N ≥ 2**
 1. `git switch <branch>` — do **not** branch off it.
 2. Implement, lint, commit, push.
-3. `gh pr edit <branch> --title "Implemented phase <ordinal>"` and
+3. `gh pr checks <branch> --watch`, and fix what it reports.
+4. `gh pr edit <branch> --title "Implemented phase <ordinal>"` and
    `gh pr comment <branch> --body "Phase <ordinal>: <what it added + how to verify>"`
-4. `promote_spec.py <slug> --to in-progress --detail "P<n> done"`
-5. **If it was the final phase and it merged:**
+5. `promote_spec.py <slug> --to in-progress --detail "P<n> done"`
+6. **If it was the final phase, CI is green and it merged:**
    `promote_spec.py <slug> --to completed --detail "all phases in <base> (PR #<n>)"`
+
+**CI is a gate, not a report.** A phase whose tests pass locally and fail on the
+runner is unfinished, and a check that was already red before the branch is still
+the branch's problem — nothing merges past it. Fix it in its own commit.
 
 Ordinals are spelled out: one, two, three, four, five.
 
@@ -200,6 +206,9 @@ completed.
 - One subfolder per feature, always inside a status bucket — never loose in
   `specs/` or in a bucket root.
 - One branch and one PR per feature. Never a second PR for a later phase.
+- **A spec reaches `completed/` only with a green CI run behind it.** The bucket
+  is a claim about shippability; a red PR filed under "completed" is the same lie
+  as a shipped feature filed under "not started".
 - Phase numbers and PR ordinals stay in lockstep: phase 1 is "Implemented phase
   one".
 - A phase with no concrete "Done when" is too big or too vague. Reslice it.
--- a/.claude/skills/issue-to-phases/assets/prompt.md.template
+++ b/.claude/skills/issue-to-phases/assets/prompt.md.template
@@ -90,6 +90,29 @@ One branch, one rolling PR, retitled each phase. Never a second PR.
 - Conventional Commits: `type(scope): summary`. Write the message with the
   `technical-writing` skill.
 
+## CI is part of the work, not a report on it
+
+Every push runs the repo's checks on the PR. **A phase is not finished until they
+are green**, and you find that out inside your own iteration rather than leaving
+it for whoever opens the PR.
+
+```bash
+gh pr checks {{BRANCH}}                     # bucket per check
+gh pr checks {{BRANCH}} --watch             # block until the run finishes
+gh run view --job <id> --log-failed         # the actual failure, not the summary
+```
+
+Three rules, each of which has cost a real iteration:
+
+- **Push, then wait.** The run takes minutes. Ending your turn on "CI is running"
+  ends it with the outcome unknown, and the loop's own gate will retract your
+  marker while you are not there to fix it.
+- **A check that was already red is yours.** "It was failing before my branch" is
+  not a defence: nothing merges past a red check, so the branch that needs it green
+  is the branch in front of you. Fix it in its own commit and say so in the message.
+- **Reproduce locally before pushing again.** The lint command in *Tests and lint*
+  is the one the runner uses. A push-and-see loop burns ten minutes per guess.
+
 ## Spec status — use the script, never edit by hand
 
 Status lives in three places that must agree: the bucket folder, the README
@@ -103,7 +126,7 @@ cd {{APP_DIR}}
 python3 {{PROMOTE}} {{SPEC_SLUG}} --to in-progress --detail "P{{PHASE_NUMBER}} started"
 # Every later phase, when it lands:
 python3 {{PROMOTE}} {{SPEC_SLUG}} --to in-progress --detail "P{{PHASE_NUMBER}} done"
-# The FINAL phase ({{PHASE_COUNT}}), once merged or the PR is ready:
+# The FINAL phase ({{PHASE_COUNT}}) — ONLY once `gh pr checks {{BRANCH}}` is all green:
 python3 {{PROMOTE}} {{SPEC_SLUG}} --to completed \
   --detail "all phases in {{BASE_BRANCH}} (PR #<n>)" --note "<one line for the index>"
 # Verify before writing your done marker:
@@ -112,6 +135,10 @@ python3 {{PROMOTE}} {{SPEC_SLUG}} --check
 
 The loop gates on this. A phase that leaves the spec in the wrong bucket has its
 done marker retracted, so skipping it costs an iteration.
+
+`completed/` means the whole feature is shippable, so a spec moves there only with
+a green CI run behind it. A red PR filed under "completed" is the same lie as a
+shipped feature filed under "not started".
 
 {{SAFETY_RULES}}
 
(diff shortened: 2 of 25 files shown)


Venkateshvenki404224:
## Phase 2 — placement, lazy bridges, the attachment reconciler, and a capacity number

A bench no longer goes to index 0 because it is the only index.

- **`placement.py`** — `pick_network()` walks the family and returns the lowest bridge under `bench_slots_per_bridge`. A bridge that does not exist counts as empty, so a small install runs on one bridge and pays for one. A full family is refused with a sentence naming the count and the cap, not a Docker error. No lock, so the `Bench Instance` → `Credit Account` → `Bench Admission` order is untouched.
- **The roll lives at start, not at create.** On Docker 29.7.2 the create succeeds and the *start* raises `no available IPv4 addresses on this network's address pools`. `start_bench_container` catches exactly that, advances one index, recreates and starts. One retry, never a loop. `tests/test_docker_manager.py` pins the wording this daemon really emits.
- **`_deploy_bench`** picks a bridge for a `Draft` bench, then re-reads `HostConfig.NetworkMode` after the start and records *that* — a measured verdict, like the runtime line beside it. The deploy log gains `bench bridge <name>`.
- **`reconcile_instance_routes`** gains an attachment pass on the same `*/5` job. It only touches bridges that already exist, and logs only what it actually reattached.
- **`run_diagnostics`** gains `bridge_capacity` — used and free per existing bridge plus total headroom. Row count is now eight in all four places that state it.
- **`scripts/density_drill.py`** fills a scratch `bpdrill-*` network until the daemon refuses. It asserts the target name before the first create, refuses a mismatched `--i-know-this-is`, and removes everything in a `finally`.

New settings: `bench_bridge_count` (16) and `bench_slots_per_bridge` (1000), mirrored in the SPA under a new **Bench network** group along with `bench_subnet_base`.

### How to verify

```bash
# 1. What a bridge really refuses at — safe, it fills a network it made
python3 scripts/density_drill.py --subnet-size 24 --endpoints 260 --i-know-this-is <base_domain>

# 2. Force a roll: cap the bridge at one bench, deploy, then put the cap back
bench --site frontend execute frappe.client.set_value \
  --kwargs '{"doctype":"BenchPress Settings","name":"BenchPress Settings","fieldname":"bench_slots_per_bridge","value":1}'
# ...deploy a bench, then:
docker network inspect benchpress-1 --format '{{range .Containers}}{{.Name}} {{end}}'
curl -sI https://<new_id>.<base_domain>/api/method/ping

# 3. Break the attachment and let the tick fix it
docker compose up -d --force-recreate traefik
docker inspect benchpress_traefik --format '{{.State.StartedAt}}'   # before and after — must not change
docker network inspect benchpress-1 --format '{{len .Containers}}'  # 3, then 4 one tick later

# 4. The capacity row must agree with the daemon
bench --site frontend execute benchpress.diagnostics.run_diagnostics
```

### Measured on staging

| Check | Result |
|---|---|
| Drill, `/28` / `/26` / `/24` | held 13 / 61 / **253**, refused at 14 / 62 / **254** — exactly `2^(32-n) - 3` each time |
| Deploy with the cap at 1 | landed on `benchpress-1`, `10.20.16.5`, device `bpbr1`, `10.20.16.0/20` |
| `benchpress-1` members | the bench plus exactly `benchpress_traefik`, `benchpress-mariadb`, `benchpress-redis` |
| Ingress to the new bridge | **200** `{"message":"pong"}` — only possible if Traefik joined a bridge that did not exist when it started |
| Traefik recreated | `benchpress-1` fell to 3 endpoints, the bench answered 502 |
| One `*/5` tick later | back to 4 endpoints, **200**, and `StartedAt` byte-identical — attached, not restarted |
| `bridge_capacity` row | `benchpress-0 4 used / 996 free, benchpress-1 4 used / 996 free — 15992 benches of headroom`, agreeing with `docker network inspect` |
| Negative control | no `benchpress_db`, `benchpress_redis-cache` or `benchpress_redis-queue` on `benchpress`, `benchpress-0` or `benchpress-1`; all three still on `frappe_network` only |

Suite: 769 tests, 7 failures, all pre-existing and all live-data artefacts — five in `test_credit_sweep` (confirmed by re-running them on the phase-1 tree), plus the `test_demo_data` seed age and `test_signup` counts phase 1 already recorded. The cap was restored to 1000 and `pick_network()` reads `benchpress-0` again.

**Known gaps, by design:** nothing rebalances, the legacy `benchpress` network keeps its original benches, and no host kernel ceiling has moved — that is phase 3.

Venkateshvenki404224:
## Phase 3 — the host's ceilings are written and verified

The last phase of `density-kernel-limits`. It turns the host's kernel ceilings from whatever the machine image shipped into a written package, and keeps "written" and "in force" as separate questions.

### What landed

**`benchpress_devops/host_tuning.py`** (new) — the knob table and two functions that are never interchangeable. `targets(benches)` says what this host should have. `measure(names)` reads `/proc/sys` and says what it has. `render(benches, bridges, output=...)` writes the drop-in for knobs measured **below** target and returns the names written. `verdict()` and `report()` compare the two, and `bench_ceiling()` reports what `kernel.threads-max` implies.

**`benchpress_devops/scripts/tune-host.sh`** (new) — the `enable-sysbox.sh` convention: repo-root `cd`, `set -euo pipefail`, `usage()`, `--yes`, `--benches N`, `--bridges N`, a root assertion carrying its reason, and a timestamped backup of any existing drop-in. It ends by re-reading `/proc/sys` and exits non-zero when a knob is still low — then greps `/etc/sysctl.d` for other files naming that key.

**`benchpress_devops/entry.py`** — a bare `--check-host` flag that prints the same table read-only. It never escalates: the read is in Python, the write stays in the script.

**`benchpress/diagnostics.py`** — a `kernel_ceilings` row reading the three knobs a container can see. Row count moves eight → nine in all four places that state it.

**`benchpress_devops/PRODUCTION.md` §5.18** — the measured table and the claims. Test count bumped 84 → 131.

### The distinction this phase is built around

A file under `/etc/sysctl.d` is a request, not a value. This host proves it: `99-ci-runner.conf` asks for `fs.inotify.max_user_instances=512` and `sysctl -n` reads `1048576`. So every number anyone reports comes from `measure()`, and nothing reports a number it wrote.

Two things deliberately not copied from the reference implementation:

- It ends with `sysctl --system >/dev/null 2>&1` and prints `[OK]` either way. The gate here re-reads `/proc/sys` and exits non-zero.
- Its numbers disagree with its own documentation in three places. Every value here derives from `--benches N`, and that arithmetic goes into the rendered file as the comment above the knob.

`kernel.threads-max` is reported, never written. It is sized from RAM, and raising it past what memory supports turns a clean refusal into an OOM kill. At 55448 threads and the 500-pid bench cap, this host tops out around **110 benches** — the ceiling neither the original item nor the reference names.

### How to verify

Read-only, no root:

```bash
cd benchpress_devops
./entry.py --check-host          # exits 1 today, naming kernel.pty.max and gc_thresh1
python3 -m unittest discover -s tests -v          # 131 pass
python3 -c 'import tempfile, pathlib, host_tuning; \
  out = pathlib.Path(tempfile.mkdtemp())/"r.conf"; host_tuning.render(1000, 2, output=out); \
  print(out.read_text())'
docker compose exec -T backend bench --site frontend execute benchpress.diagnostics.run_diagnostics
```

The rendered file contains `kernel.pty.max = 9024`, `gc_thresh1 = 4096`, `netdev_max_backlog = 16384`, `tcp_max_syn_backlog = 8192`, and dash-prefixed `-net.ipv4.conf.bpbr<i>.proxy_arp = 1`. It contains **no** `kernel.pid_max` line — that knob is already at target here, and restating the cloud image's values would make this repo the owner of numbers it did not choose.

The `-` prefix is load-bearing. Bench bridges are created lazily, so most do not exist at boot, and systemd-sysctl reads `-` as "apply if the device exists, do not fail otherwise". A bridge created months from now still gets the setting: udev's `99-systemd.rules` re-runs `systemd-sysctl --prefix=/net/ipv4/conf/$name` when a network device is added.

**Applying it is a human step and was not run here.** The Ralph loop that implemented this phase gates on `/etc/sysctl.d/99-benchpress-density.conf` *not* existing. Changing kernel ceilings host-wide from an unattended session is the one failure mode nothing downstream could recover from. To finish:

```bash
sudo ./scripts/tune-host.sh --benches 1000 --bridges 2 --yes
sysctl -n kernel.pty.max                      # 9024
sysctl -n net.ipv4.conf.bpbr0.proxy_arp       # 1
sysctl -n net.ipv4.conf.all.proxy_arp         # 0 — per bridge, never for `all`
./entry.py --check-host                       # now exits 0
```

Undo is one file: `sudo rm /etc/sysctl.d/99-benchpress-density.conf && sudo sysctl --system`.

### State on the staging host

131 devops tests pass. The app suite is 769 + 121 with the same 7 failures phase 2 recorded, all pre-existing and all on live data. Lint clean apart from the known `promote_spec.py` trap. All 13 compose services up, the control plane answers pong, and all three benches — one on `benchpress`, one on `benchpress-0`, one on `benchpress-1` — still serve 200. `/etc/sysctl.d` is byte-identical to the preflight snapshot.

One unrelated change rides along: `scripts/density_drill.py` had drifted from `ruff-format` since phase 2, and the formatter fixed it.

PR #165, opened 2026-08-26
Implemented phase three

Venkateshvenki404224:
Phase 1 of the golden-bench-images spec: a lab image carries its finished
site's database, and a spawn restores it instead of installing it.

## What is here

- `benchpress/golden.py` — `build_golden` starts a throwaway container from the
  lab's existing image, creates the site once with the same `setup-site.sh` a
  deploy runs (`USE_GOLDEN=0`), dumps the database through gzip, and appends one
  layer over the same tag carrying `/opt/benchpress/golden/{site.sql.gz,manifest.json}`
  and the script that reads them. The image is never rebuilt. The golden
  database and the container are dropped in a `finally`.
- `benchpress/api.py` — `build_lab_golden`, System-Manager-only, queues it on
  `queue-long` with a `Build Log`.
- `setup-site.sh` — a third branch between adopt and create: `bench new-site
  --source-sql` plus `set-admin-password`, because every tenant of a lab
  restores the same bytes.
- `deploy_manager.py` — `GOLDEN_MARKER`, and step 7 now reports which of the
  three branches ran.

## Measured on staging (CRM lab, `benchpress/crm:lab`)

| | before | after |
|---|---|---|
| site step | 40.4 s | **9.0 s** |
| whole deploy | 44.0 s | **14.1 s** |
| image growth | — | 266 KB over 9.1 GB |

Three changes were needed to get the site step under the spec's ten seconds,
all inside its own budget and all measured:

- the dump skips per-table `DROP TABLE`, `LOCK TABLES` and `DISABLE KEYS`. The
  restore is DDL-bound — a schema-only dump of the same site restores in the
  same time as the full one — so those statements are pure overhead against the
  empty database `bench new-site` has just created. 8 s to 6 s.
- `execute_sql` is one `docker exec` instead of three, and a sequence of
  statements is one call instead of one per statement. Twelve calls became
  three, ~2 s.
- `developer_mode` and `default_site` come from `common_site_config.json`, which
  the control plane already writes at step 6. `bench use` was writing the same
  `default_site` key. Two `bench` invocations, ~1 s.

The previous site's database is dropped beside the previous container now,
in the container step, where the rest of a redeploy's teardown already happens.

## How to verify

```bash
# 1. Build the golden for a lab whose image exists.
docker compose exec -T backend bench --site frontend execute \
  benchpress.api.build_lab_golden --kwargs '{"lab_name":"crm"}'

# 2. The artefact is in the image, not on a host path.
docker run --rm --entrypoint bash benchpress/crm:lab -c \
  'ls -l /opt/benchpress/golden/ && zcat /opt/benchpress/golden/site.sql.gz | head -3'

# 3. Nothing leaked.
docker ps -a --filter name=bpgolden- --format '{{.Names}}'
docker exec -e MYSQL_PWD=$ROOT_PW benchpress-mariadb \
  mariadb -u root -N -B -e "SHOW DATABASES LIKE '\_bpgolden%'"

# 4. Deploy a bench of that lab; step 7 says "restored from the image's golden
#    dump" and the gap to step 8 is under ten seconds.
# 5. Inside the container: `bench --site <site> list-apps` lists frappe and crm,
#    and the lab page's password logs in over HTTPS.
# 6. A lab with no golden still deploys and its log says the site was created.
```

All six were run against staging. 5: `frappe 15.118.0` + `crm 1.81.2`, 281
tables, `POST /api/method/login` returns `Logged In` and `/crm` returns 200.
6: the `erpnext` lab deployed in 91.1 s with an 88.6 s site step and
"Site created successfully".

`vpn-management-16` is not usable as the negative control the spec names: its
app's `install-app` needs the control plane's `wg-agent` socket, which no lab
container has, so a bench of it has never deployed on this host. That is
pre-existing and unrelated to this branch.

## Known gaps, left to later phases

The golden is produced by hand for one lab (phase 2), nothing records which labs
have one (phase 2), the dump is appended without being restored anywhere first
(phase 3), and a golden taken from a different MariaDB major version would still
be restored (phase 3).

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -160,7 +160,29 @@ Layer 5: Create site (create-site.sh)
 
 **Container entrypoint** (`entry.sh`): Starts MariaDB, Redis, SSH, then `tail -f /dev/null` to keep alive.
 
-**Post-deploy site creation** (`setup-site.sh`): Used when creating additional sites in a running container.
+**Site creation on deploy** (`setup-site.sh`): the deploy execs this inside the new container,
+and it takes one of three paths.
+
+| Path | When | Cost on a CRM lab |
+|------|------|-------------------|
+| **Adopt** | The container already has `sites/<site>/site_config.json` — a redeploy that kept its data | Re-sets the admin password, nothing else |
+| **Restore** | The image carries `/opt/benchpress/golden/site.sql.gz` and the deploy allows it | `bench new-site --source-sql`, **9.1 s** |
+| **Create** | Anything else | `bench new-site` plus `install-app` per app, **37.2 s** |
+
+The golden dump is baked by `benchpress/golden.py` at the end of every image build: it runs the
+lab's site once in a throwaway container, dumps that database, restores the dump into a scratch
+database to prove it before trusting it, and appends one `FROM <tag>` + `COPY` layer to the tag
+the build just produced. It never rebuilds the image.
+
+Two switches in **BenchPress Settings** govern it, and they are separate on purpose:
+`enable_golden_images` decides whether a build bakes one, `restore_from_golden` whether a deploy
+reads one. A host that turns baking off keeps restoring from the images it already has.
+
+`deploy_manager._golden_matches_server` is the last gate. The dump is the one artefact in an
+image whose validity depends on something outside it, so a deploy compares the MariaDB **major**
+version the dump came from against the server it is restoring into and takes the create path
+with the reason in the log when they differ. A patch-level difference is not a mismatch —
+refusing on one would take every golden on the host out of service on a routine server update.
 
 ---
 
--- a/README.md
+++ b/README.md
@@ -296,6 +296,7 @@ flowchart LR
 - **Resource Controls** -- CPU cores and memory limits per lab, enforced by Docker `--cpus` and `--memory` flags
 - **Container Management** -- Start, stop, restart, redeploy, and delete benches from the dashboard
 - **Automatic Site Provisioning** -- Each bench gets its Frappe site created automatically on deploy, with the lab's apps installed
+- **Golden Images** -- A lab's build bakes the finished site's database into the lab's own image, so a deploy restores it instead of creating 281 tables through the ORM. Measured on a 2-vCPU host, a CRM lab's site step goes from **37.2 s to 9.1 s** and the whole deploy from 43.3 s to 13.3 s. Re-run it yourself with `python3 scripts/golden_drill.py --lab crm --runs 3 --i-know-this-is <your base domain>`, and add `--cold` for the control. A lab with no golden still deploys -- it is just slower
 - **VPN Device Management** -- Register persistent devices (Laptop, Mobile, etc.), generate WireGuard configs per device, and manage device lifecycle from a dedicated page
 - **Confirmation Dialogs** -- Destructive actions (deploy, stop, delete) require explicit confirmation before execution
 - **Stats Monitoring** -- CPU and memory usage polled every minute from the Docker stats API, displayed as progress bars
--- a/benchpress/api.py
+++ b/benchpress/api.py
@@ -96,6 +96,26 @@ def build_lab_image(lab_name: str) -> dict:
 
 
 @frappe.whitelist()
+def build_lab_golden(lab_name: str) -> dict:
+	"""Queue the golden build for one lab, appending it to the image the lab already has."""
+	require_admin()
+	lab = frappe.get_doc("Lab", lab_name)
+	tag, hit = image_cache.resolve(lab)
+	if lab.status != "Ready" or not hit or lab.image_tag != tag:
+		frappe.throw(_("No built image for lab '{0}'. Build it first from the Lab record.").format(lab.title))
+	frappe.enqueue(
+		"benchpress.golden.build_golden_job",
+		lab_name=lab_name,
+		user=frappe.session.user,
+		queue="long",
+		timeout=image_cache.BUILD_TIMEOUT,
+		job_id=f"golden:{lab_name}",
+		deduplicate=True,
+	)
+	return {"name": lab_name, "status": "Queued"}
+
+
+@frappe.whitelist()
 def prewarm_catalog() -> dict:
 	"""Build the shared image for every catalog template that has none yet.
 
--- a/benchpress/benchpress/doctype/benchpress_settings/benchpress_settings.json
+++ b/benchpress/benchpress/doctype/benchpress_settings/benchpress_settings.json
@@ -10,6 +10,7 @@
   "docker_socket",
   "default_image",
   "default_bench_runtime",
+  "enable_golden_images",
   "column_break_docker",
   "base_domain",
   "traefik_network",
@@ -50,6 +51,13 @@
    "fieldtype": "Select",
    "label": "Default Bench Runtime",
    "options": "runc\nsysbox"
+  },
+  {
+   "default": "1",
+   "description": "Bake a lab's finished site into its image on every build, so a deploy restores the database instead of creating it. Turning it off shortens builds on a disk-constrained host; images that already carry a golden keep restoring from it.",
+   "fieldname": "enable_golden_images",
+   "fieldtype": "Check",
+   "label": "Enable Golden Images"
   },
   {
    "fieldname": "column_break_docker",
--- a/benchpress/benchpress/doctype/benchpress_settings/benchpress_settings.json
+++ b/benchpress/benchpress/doctype/benchpress_settings/benchpress_settings.json
@@ -11,6 +11,7 @@
   "default_image",
   "default_bench_runtime",
   "enable_golden_images",
+  "restore_from_golden",
   "column_break_docker",
   "base_domain",
   "traefik_network",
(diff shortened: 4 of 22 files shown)


Venkateshvenki404224:
Phase 1 verified on staging against the live host.

**The golden is in the image.** `benchpress/crm:lab` carries `/opt/benchpress/golden/site.sql.gz` (266 099 B, 277 `CREATE TABLE`) and `manifest.json`. The previous image kept its layers and became the parent of the new tag — nothing was rebuilt, nothing was removed, and free disk is unchanged at 21 G.

**A redeploy restores instead of installing.** Bench `c663148574f4a57aab5a6422b69abccf`, crm lab:

```
=== Step 7/11: Creating the site [site @4.4s] ===
Site c663148574f4a57aab5a6422b69abccf.benchpress.cloud restored from the image's golden dump
=== Step 8/11: Preparing assets [assets @13.1s] ===
```

**8.7 s** for the site step and **13.9 s** for the whole deploy, against 40.4 s and 44.0 s cold.

**The restored site is a working site.** `bench list-apps` reports `frappe 15.118.0` and `crm 1.81.2`; `POST /api/method/login` with the password the lab page shows returns `Logged In`; `GET /crm` serves 200; `frappe.client.get_count` answers for `CRM Lead` and `CRM Deal`.

**A lab with no golden still deploys.** A fresh `erpnext` bench took the cold path — `site @1.9s` → `assets @90.3s`, log line `Site created successfully` — and was deleted afterwards. (The spec named `vpn-management-16` for this, but that lab cannot deploy on this host for an unrelated reason: `vpn_management.install.after_install` health-gates on a `wg-agent` socket that does not exist inside a lab container.)

**Nothing leaked.** No `bpgolden-` container, no `_bpgolden%` database in the shared MariaDB, and every lab image tag present before the run is still present.

**Tests.** `benchpress.tests.test_golden` — 8 tests, OK. The full app suite fails the same eight live-data tests on this branch and on `version-16`, so no regression; CI runs them on a clean site and is green.

### How to verify

```bash
docker run --rm --entrypoint bash benchpress/crm:lab -c \
  'ls -l /opt/benchpress/golden/ && cat /opt/benchpress/golden/manifest.json'

cd /home/ubuntu/benchpress_devops && </dev/null docker compose exec -T backend \
  bench --site frontend execute benchpress.api.build_lab_golden --kwargs "{'lab_name':'crm'}"
```

Then redeploy a bench of that lab and read step 7 of its Deploy Log.

Venkateshvenki404224:
## Phase two — a golden on every build, and a catalog that says who has one

Phase one made a golden something an admin ran by hand. This phase makes it a property of a build.

**What it adds**
- `BenchPress Settings.enable_golden_images` (default on). A disk-constrained host turns the extra build step off; images that already carry a golden keep restoring from it.
- `_build_lab_with_logs` runs the golden step after the lab reaches `Ready` with its tag saved — same job, same Build Log. It **never raises**: a lab whose image built and whose golden did not is a working lab with a slow deploy.
- The appended layer stamps `benchpress.golden` and `benchpress.golden.mariadb`. A deploy reads those labels, never `Lab.golden_manifest` — the row is a claim, the image is the artefact.
- `Lab.golden_manifest` records what the last build baked, or nothing when it baked nothing. Also cleared on a spec change and when the image tag moves.
- A deploy against an image with no golden now names the action that fixes it.
- `prewarm_catalog` bakes a golden in the same job it builds a template in.
- `run_diagnostics` gains a `Golden images` row at **Warning** severity — a missing golden is slow, not broken.

**Verified live on staging**

| Check | Result |
|---|---|
| Build end to end (`erpnext`) | One Build Log: image build → `Baking the golden site` → scratch container → dump (783 525 B, 702 tables) → `LABEL benchpress.golden="1"` → manifest |
| Row and image agree | `benchpress.golden=1`, `benchpress.golden.mariadb=10.6.28-MariaDB-ubu2204`; `Lab.golden_manifest` populated |
| Deploy restores | erpnext site step **92.9 s cold → 17.8 s golden**; crm **9.6 s**. Restored site: `frappe 15.119.0`, `erpnext 15.120.0`, `/login` 200, `POST /api/method/login` → `Logged In` with the deploy's own password |
| Switch off | Build Log says `Golden images are turned off`, no `/opt/benchpress/golden` in the image, no label, empty manifest — and the deploy still succeeds, logging `No golden dump in benchpress/erpnext:lab` |
| Failed golden | Deliberate raise: Build Log records `Golden step failed: … the image is usable and deploys will be slow`, `log_type` stays `success`, `Lab.status` stays `Ready` with its tag |
| Diagnostics | 10 rows; `golden_images` = `2 of 8 built labs carry a golden dump`, severity `Warning`, missing tags named |
| Nothing leaked | No `bpgolden-` container, no `_bpgolden%` database, all 8 lab tags intact, disk unchanged |

**One thing worth knowing:** phase one's golden on `benchpress/crm:lab` predates the labels, so it read as missing. Any image goldened before this phase needs `build_lab_golden` re-run on this code, or the catalog understates coverage. crm was re-goldened here.

Full suite: 778 tests, the same 7 failures + 1 error as `version-16` on this host (they read live staging rows); CI's clean-site run is green. `uvx pre-commit@4.3.0 run --all-files` passes.

Venkateshvenki404224:
## Phase 3 — a golden is proven before a tenant meets it, and the win is measured

Two things stop being assumptions.

**That the dump restores.** `golden._verify_dump` runs between the dump and the appended layer: it creates `_bpgolden_verify_<lab_id>`, pipes the dump back in through `gzip -cd` exactly as `bench new-site --source-sql` will, counts `information_schema.tables` and reads back `tabInstalled Application`. A dump that returns no tables, or fewer apps than the lab installs, is never baked — the image keeps the cold path, which works. `tables`, `installed_apps` and `restore_seconds` go into the manifest. The scratch database and its limited user are dropped in a `finally`.

**That it is faster.** `scripts/golden_drill.py` deploys one lab through the shipped `create_bench` endpoint, waits for `Running`, and reads each run's site step out of its own Deploy Log via `deploy_pipeline.parse_step_line`. `--cold` is the control and it is the same image.

Plus the check that cannot be made inside the image: `deploy_manager._golden_matches_server` compares the MariaDB **major** version the dump came from with the server it would be restored into. A patch bump is not a mismatch — refusing on one would take every golden on the host out of service the next time the server is updated.

### One decision worth flagging

Phase 3 adds a **second** switch, `BenchPress Settings.restore_from_golden`, rather than overloading `enable_golden_images`. The spec's own confirmed decision is that turning baking off must not stop images that already carry a golden from restoring, so one flag cannot mean both. The new one is what a deploy reads, what the drill moves to measure its cold arm on the same image, and what takes a suspect golden out of service without a rebuild.

Both are `Check` fields, and a Frappe Single stores nothing until something writes it — so on an untouched site `enable_golden_images` read as `None`, i.e. **off**, and builds baked no golden silently. `patches/default_golden_switches.py` writes both defaults. That gap shipped in phase 2; it is fixed here.

### Verified live on staging

| # | Check | Result |
|---|---|---|
| 1 | The verification restore runs and is recorded | `benchpress/crm:lab` manifest carries `tables: 281`, `installed_apps: ["crm","frappe"]`, `restore_seconds: 6.0`. Build Log: `Restored 281 tables and crm, frappe in 6.0s` |
| 2 | A bad dump is never appended | Dump truncated to 100 bytes on purpose: `=== Build failed: Golden dump would not restore (exit 1): gzip: unexpected end of file ===`. Image id **byte-identical** before and after (`sha256:6029dfaa…`), so crm kept its previous good golden. No `_bpgolden%` database and no `bpgolden-` container survived |
| 3 | The version gate fires — and only on a major | Scratch tag labelled `11.4.5-MariaDB`: `Golden dump was taken from MariaDB 11.4 and this server is 10.6 — creating the site instead`, site step **39.3 s** (cold). Same tag relabelled `10.6.31-MariaDB` against a `10.6.28` server: **restored**, site step **9.2 s**. Scratch bench, lab and tag removed; the lab-image tag set is identical to the pre-run snapshot |
| 4 | The drill reports both arms | `--cold`: n=3, p50 **37.2 s**, p95 37.6 s, deploy p50 43.3 s. Golden: n=3, p50 **9.1 s**, p95 9.2 s, deploy p50 **13.3 s**. Both match the spec's facts table (36.5–43.0 s cold, restore in 6.6 s) |
| 5 | The drill left nothing behind | No `golddrill` container, no `Bench Instance` or `Bench Site` row, no `_bpgolden%` database, `restore_from_golden` back at `1` |
| 6 | Tests and lint | `run-tests --module benchpress.tests.test_golden` → 45 tests, OK. Full app suite: the same 8 pre-existing failures as the `version-16` baseline (`test_credit_sweep` ×5, `test_lease`, `test_signup`, `test_demo_data`), which read live staging rows. `uvx pre-commit@4.3.0 run --all-files` clean |

### Reproduce it

```bash
python3 scripts/golden_drill.py --lab crm --runs 3 --i-know-this-is <your base domain>
python3 scripts/golden_drill.py --lab crm --runs 3 --cold --i-know-this-is <your base domain>
```

The cold arm moves `restore_from_golden` for the run and puts it back in a `finally`, the same way `admission_drill.py` moves the concurrency cap.

### Note for a reviewer

The spec's verification step 5 greps for `golddrill` in container names. It cannot match: `BenchInstance.before_insert` names every bench `md5(email + lab)`, so a caller cannot choose a container name. The drill's benches are instead identified by their owner, `golden-drill@example.com` — a user only `benchpress/golden_drill.py` mints — and by the site label `golddrill-<lab>`. Cleanup filters on the owner, never on status.

Issue #176, opened 2026-08-26
ADR: dial-out control bus with signed commands (item 11)


Venkateshvenki404224:
## Question

Write the ADR for the dial-out control bus (todo item 11).

Decided now, built later, and deliberately so: nothing in the codebase has a node or host abstraction — deployment is single-host against a local Docker socket — so there is nothing to attach a bus to yet. The reason to decide now is that anything built before it must not assume inbound access, and that constraint is free today and expensive later.

The item's position, to be accepted, rejected or amended:

- Nodes dial **out** to a Redis/Valkey Streams broker. No inbound ports on customer boxes.
- JSON payloads only.
- Heartbeat as a TTL key.
- Per-site locks with preempting verbs, so `site.stop` can preempt a stuck migrate.
- Reclaim with an age ceiling. Their broker replayed a backlog of stale `vm.delete` commands the moment reclaim started working — the age ceiling is the fix for exactly that.
- The agent verifies a signature on every command before dispatch. Their unsigned bus is the finding their own `SECURITY.md` says blocks real customers.

Settle at minimum: what signs and what verifies, where the key lives, what happens to a command that fails verification, what the age ceiling is measured against given clock skew between node and broker, and which verbs preempt which.

Per the map's destination, this produces an **ADR** under `specs/decisions/`, not a phase spec — the alternatives considered, the decision, and the constraints it places on work built before it. The last part is what the other tickets consume.

**Resolve with the ADR written and linked from this ticket.**

Venkateshvenki404224:
## Out of scope

**The dial-out control bus is not built and its ADR is not written.** Item 11 designs a command channel to a fleet BenchPress will not have. Item 12 goes with it, in [ADR: capacity accounting (item 12)](https://github.com/Venkateshvenki404224/benchpress/issues/177).

This is a scoping act, not a step on the route, so it is not in Decisions so far. Everything below is why, and the one sentence that survives.

### The bus already exists, and it already dials out

Item 11 reads as "add a bus where none exists". Four mechanisms say otherwise, and none of them opens an inbound port:

- **Workers already phone home.** Every bench-running process is a Frappe worker container that connects *outward* to Redis and MariaDB. Nothing connects inward. Dial-out is not a proposal here; it is the only thing that has ever been built.
[...]

Issue #166, opened 2026-08-26
Map: scaling BenchPress for the agent-runtime load profile (todo items 7-19)


Venkateshvenki404224:
## Destination

Phase specs under `specs/` for items 7-10 and 13-19 of `todo.md` ("Scaling imports from labs-devops"), each written so an unattended Ralph loop can build it. The map is done when every one of the eleven items has its decisions closed and nothing is left to settle before spec-writing starts.

Items 11 and 12 were charted here and are now **out of scope, ADRs included** — the destination shrank from thirteen items to eleven. See below.

## Notes

**Domain.** BenchPress: a Frappe app that builds Docker images and spawns per-tenant bench containers behind Traefik. The load profile driving all thirteen items is the agent runtime — spawn, work, destroy, hundreds of times a day. Items 1-5 of `todo.md` already shipped in this shape; `specs/completed/` holds fifteen efforts to copy from.

**Skills every session should consult.** `frappe-app-dev` for anything touching doctypes, controllers, whitelisted APIs, jobs or permissions. `code-style` for any code change. `dsa-conscious-coding` alongside it for build/deploy orchestration, container lifecycle and credit-ledger work. `technical-writing` for the spec prose and commit messages. `issue-to-phases` once an item's decisions are all closed. `quality-code-review` on any diff.

**Standing preferences for this effort.**
- **Executor is an unattended Ralph loop.** A ticket is not resolved until a loop could act on its answer without asking a follow-up question. This is a higher bar than "a human would understand it".
- **Spec-writing is the handoff, not a route step.** Tickets resolve decisions. When every decision for an item is closed, that item's spec folder is written with `issue-to-phases`. Do not write specs from inside a ticket.
- **Host halves are in scope.** Several items need work in the parent `benchpress_devops` repo. Its changes belong in the same spec, in a section of their own, and any step no loop may run (sudo, `daemon.json`, a reboot) is written as an explicit human checklist. Item 5 is the precedent: its spec ends "Outstanding, by hand: `sudo benchpress_devops/scripts/tune-host.sh`."
- **Item 6 (golden bench images) is an external prerequisite, and it has landed.** Merged 2026-08-26 as `d58b64b` on `version-16` (PR #165). It is not charted here. Item 7's healthcheck start period is the one place its landing changes an answer, and that answer is now measurable rather than assumed.
- **CI is a gate.** Nothing is done until the PR's checks are green.
[...]

PR #188, opened 2026-08-26
docs(decisions): add ADR for the dial-out control bus

Venkateshvenki404224:
Resolves the wayfinder ticket [ADR: dial-out control bus with signed commands (item 11)](https://github.com/Venkateshvenki404224/benchpress/issues/176), a child of [Map: scaling BenchPress for the agent-runtime load profile](https://github.com/Venkateshvenki404224/benchpress/issues/166).

## What this adds

One file: `docs/decisions/0001-dial-out-control-bus.md`. No code changes.

The ADR accepts todo item 11. Nodes dial out to a Redis or Valkey Streams broker, commands are JSON, and every command carries an Ed25519 signature that the node verifies before dispatch. It is accepted and not built, because no node abstraction exists yet.

## Why the file is here and not in `specs/decisions/`

The map's destination said `specs/decisions/`. `.gitignore:71` hides `specs/`, `git ls-files specs/` returns zero files, and a completed spec is deleted once its code lands. An ADR whose job is to constrain work built later cannot be disposable, so it goes in tracked `docs/`.

## Findings that shaped it

- Frappe builds every RQ queue with no `serializer` argument, so RQ 2.6.1 applies its pickle default. Measured: `resolve_serializer(None).dumps({'a': 1})` returns `b'\x80\x05\x95...'`. The per-node queues in `lease.stop_queue_for` are therefore a local mechanism and cannot become the bus.
- WireGuard already removes the need for a public inbound port, so a broker has to earn its place on durability rather than on reachability. The alternatives section records why it does.
- The age ceiling reads the broker clock only. Measured on `redis-queue` 6.2.23: entry ID `1787739456450-0` against broker `TIME` `1787739456000` ms, and `XPENDING` idle of `2004` ms after a two second wait.

## Review focus

The **Constraints on work built before the bus** section. Those seven lines are what the rest of the map consumes, and each one is meant to be checkable in a review.

### DIFF
--- a/docs/decisions/0001-dial-out-control-bus.md
+++ b/docs/decisions/0001-dial-out-control-bus.md
@@ -1,0 +1,233 @@
+# 1. Dial-out control bus with signed commands
+
+- **Status:** Accepted. Not built.
+- **Date:** 2026-08-26
+- **Decides:** `todo.md` item 11
+- **Ticket:** [ADR: dial-out control bus with signed commands (item 11)](https://github.com/Venkateshvenki404224/benchpress/issues/176)
+- **Map:** [Map: scaling BenchPress for the agent-runtime load profile](https://github.com/Venkateshvenki404224/benchpress/issues/166)
+
+## Context
+
+BenchPress deploys to one host. `docker_manager` talks to a local Docker
+socket, and no record anywhere names a host. A second host is not planned for
+any current release.
+
+The reason to decide now is cost asymmetry. Code that assumes it can open a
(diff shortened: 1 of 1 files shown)


Venkateshvenki404224:
Closing unmerged. [ADR: dial-out control bus with signed commands (item 11)](https://github.com/Venkateshvenki404224/benchpress/issues/176) was ruled out of scope in a parallel session while this was being written, and the map's destination shrank from thirteen items to eleven. The ADR documents a decision the map no longer wants made. Full retraction on the ticket.

PR #207, opened 2026-08-26
Implemented phase two

Venkateshvenki404224:
Phase 1 of `secrets-off-the-exec-stream`.

`docker_manager.write_file_to_container` no longer builds a heredoc. The file
content travels in the exec environment, which the Docker daemon does not
publish; the command line carries the path and the mode only. Two `chmod` execs
go with it, because the write now applies the mode itself.

**`put_archive` — the planned transport — does not work on this host.** Bench
containers run `sysbox-runc`, which re-mounts the container root as its own
overlay over the snapshot Docker uploads into. Measured on a fresh container per
runtime, same image:

| Extraction | sysbox-runc | runc |
|---|---|---|
| `put_archive('/', 'etc/wireguard/wg0.conf')` | returns `True`, file absent inside | lands |
| `put_archive('/etc/wireguard', 'direct.conf')` | returns `True`, file absent inside | lands |
| `put_archive('/tmp', ...)` | lands | lands |

The first deploy on that design failed with ``wg-quick: `/etc/wireguard/wg0.conf'
does not exist``. `mariadb_manager.py` keeps its `put_archive` because
`benchpress-mariadb` is a runc container.

## How to verify

Captured `docker events` across a live deploy of bench
`01a5ba367a13e367fe64b5e5befc07c4`, which reached Running:

| Check | Result |
|---|---|
| `exec_create` events (control — the capture was live) | 46 |
| The exact WireGuard private key that landed in the container | 0 hits |
| The exact code-server password that landed in the container | 0 hits |
| `PrivateKey` / `WGEOF` anywhere in the stream | 0 |
| The bench SSH password (phase 2 owns it) | 2 hits, as expected |

What the daemon did publish:

```
exec_create: bash -c mkdir -p /etc/wireguard && printf %s "$BENCHPRESS_FILE_CONTENT" > /etc/wireguard/wg0.conf && chmod 600 /etc/wireguard/wg0.conf
```

Files landed with the modes and owners they had before, and the two the phase
tightens:

```
600 root:root                           /etc/wireguard/wg0.conf
600 administrator:administrator         /home/administrator/.config/code-server/config.yaml
644 administrator:administrator         .../sites/common_site_config.json
775 root:root                           /opt/benchpress/scripts/linkuser.sh
```

`wg show` reports the tunnel up. The IDE returns 302 with a session cookie for
its password and 200 for a wrong one.

Tests: `test_docker_manager`, `test_mariadb_manager`, `test_vpn_adapter`,
`test_deploy_manager` all exit 0. `uvx pre-commit@4.3.0 run --all-files` clean.

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -342,3 +342,13 @@ static markup of a session that is not running anywhere.
   socket path (`unix:///var/run/docker.sock`), read at
   [docker_manager.py:132](benchpress/docker_manager.py#L132). Pointing it at a `tcp://` address on
   another host is the one edit that breaks the rule.
+- **No secret may reach a `docker exec` command line, a healthcheck command line, or a file
+  written through an exec.** Pass it as `environment=`, send a hash the server accepts, or upload
+  it with `put_archive`. Docker publishes every exec command into its event stream in full and
+  untruncated, and publishes no environment at all, so anything on that line is readable by
+  anything holding the socket. A sentinel test guards each site
+  ([test_docker_manager.py](benchpress/tests/test_docker_manager.py),
+  [test_vpn_adapter.py](benchpress/tests/test_vpn_adapter.py),
+  [test_mariadb_manager.py](benchpress/tests/test_mariadb_manager.py),
+  [test_deploy_manager.py](benchpress/tests/test_deploy_manager.py)). Base64 is not a defence:
+  `mariadb_manager.execute_sql` encodes its script and the encoding is on the line too.
--- a/benchpress/deploy_manager.py
+++ b/benchpress/deploy_manager.py
@@ -876,9 +876,9 @@ def _start_code_server(bench, container_id: str, pipeline, settings) -> None:
 	"""Configure code-server, launch it, and only then claim it is up.
 
 	Every exec is checked. The config write decides whether code-server can authenticate at
-	all, the `chown`/`chmod` decides whether it reads that file or leaks the password to every
-	account in the container, and `restart.sh` is what actually launches it — a deploy that
-	reports a `code_server_url` answering nothing is worse than one that fails here.
+	all, the `chown` decides whether it reads that file, and `restart.sh` is what actually
+	launches it — a deploy that reports a `code_server_url` answering nothing is worse than
+	one that fails here.
 
 	The address is cleared before the attempt and stored after it, so a failed launch leaves
 	the field empty and `LabHeader.showCodeServer` hides the button instead of offering a
@@ -891,10 +891,12 @@ def _start_code_server(bench, container_id: str, pipeline, settings) -> None:
 	config_yaml = f"bind-addr: 0.0.0.0:8080\nauth: password\npassword: {code_server_password}\ncert: false\n"
 	config_path = f"{cs_home}/.config/code-server/config.yaml"
 
-	write_file_to_container(container_id, config_yaml, config_path)
+	write_file_to_container(container_id, config_yaml, config_path, mode=0o600)
+	# The tar header set the mode, but `linkuser.sh` minted the tenant account and this
+	# caller does not know its id, so the ownership fix still runs as its own exec.
 	_checked_exec(
 		container_id,
-		f"chown -R {cs_user}:{cs_user} {cs_home}/.config && chmod 600 {config_path}",
+		f"chown -R {cs_user}:{cs_user} {cs_home}/.config",
 		"Securing the code-server config",
 	)
 	_checked_exec(container_id, "bash /opt/benchpress/scripts/restart.sh", "restart.sh")
--- a/benchpress/deploy_manager.py
+++ b/benchpress/deploy_manager.py
@@ -505,18 +505,20 @@ def _cleanup_failed_deploy(bench, container_id, append_log) -> None:
 	bench.wg_ip = None
 
 
-def build_linkuser_args(bench, lab, settings, ssh_password: str) -> list[str]:
+def build_linkuser_args(bench, lab, settings) -> list[str]:
 	"""Positional arguments for linkuser.sh, in the order the script reads them.
 
 	Order must match scripts/linkuser.sh:
-	USERNAME EMAIL LAB_NAME WG_IP SSH_PASSWORD BENCH_NAME BASE_DOMAIN LOGIN_SHELL
+	USERNAME EMAIL LAB_NAME WG_IP BENCH_NAME BASE_DOMAIN LOGIN_SHELL
+
+	The SSH password is not among them. It travels in the exec environment, which Docker
+	does not publish, while it publishes every command line in full.
 	"""
 	return [
 		bench.ssh_username,
 		bench.owner,
 		lab.title,
 		bench.wg_ip or "0.0.0.0",
-		ssh_password,
 		bench.bench_name,
 		settings.base_domain or "localhost",
 		lab.shell or "/bin/bash",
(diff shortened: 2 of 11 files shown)


Venkateshvenki404224:
## Phase 2 — the last three secrets leave argv

Docker publishes every exec command line into its event stream, in full and untruncated. Phase 1 took the file contents off it. This takes the remaining three secrets off.

- **MariaDB user password** — `create_mariadb_user` sends `IDENTIFIED VIA mysql_native_password USING '<hash>'` instead of the plaintext. `execute_sql` base64-encodes its script, which is not a defence, so the plaintext was published on every site creation. The bench still logs in with the plaintext; the hash is not login-equivalent, because the server hashes what a client sends once more before comparing.
- **Bench SSH password** — out of `linkuser.sh`'s argv and into the exec environment, which Docker does not publish. `build_linkuser_args` returns seven arguments; the script reads `$SSH_PASSWORD`.
- **MariaDB root password in `setup.sh`** — the readiness loop uses the image's own `healthcheck.sh --connect --innodb_initialized`, which reads its credentials from a root-owned file inside the container. The read of that password out of `.env` goes with it.

`ARCHITECTURE.md` carries the rule, and two sentinel tests bring the set to five — one per secret, each failing if someone puts it back.

### How to verify

Run the scoped tests and read `$?`, not the last line:

```
```

A test asserts on a mock, so the real proof is a capture across a live deploy. Measured here on bench `01a5ba36…`, `docker events --format '{{.Action}}'` across a full 11-step deploy:

| Check | Result |
|---|---|
| `exec_create` events — the control | **68** |
| The bench's SSH password, read back out of the database | **0 hits** |
| Its code-server password | **0 hits** |
| Its WireGuard private key, read out of `wg0.conf` | **0 hits** |
| The bench name — a second control, a string that *is* on the stream | 10 hits |
| `IDENTIFIED BY` in the decoded base64 SQL | **0** |
| `IDENTIFIED VIA mysql_native_password USING '*AE5D…'` | present |

Grepping the raw capture for `IDENTIFIED BY` proves nothing, because `execute_sql` base64-encodes the statement. Decode the payloads out of the `Action` lines first.

The linkuser exec now reads:

```
exec_create: bash -c bash /opt/benchpress/scripts/linkuser.sh administrator Administrator 'Frappe CRM' 172.27.0.8 01a5ba36… benchpress.cloud /bin/bash
```

The bench still works, which is what the hash change risks — a wrong hash fails later as a site that cannot reach its database, not at create time:

- The deploy's golden-dump restore authenticated **with the plaintext** against a user created **from the hash**.
- Site returns 200 over the public hostname and over the tunnel; `frappe.ping` answers; `frappe.db.count("DocType")` returns 318 through the site's own database user.
- SSH accepts the stored password and returns `administrator` plus the bench hostname. A wrong password is denied — the control.
- `docker exec benchpress-mariadb healthcheck.sh --connect --innodb_initialized` exits 0 on the live container.

### Still exposed, and not this branch's to fix

`DB_ROOT_PASSWORD` is in the `benchpress_db` healthcheck command and has already leaked. Rotating a live database password takes every bench off its database, because each holds the old value in its site config, so it needs a coordinated redeploy by a human.

Venkateshvenki404224:
Follow-up on phase two, commit `9568618`.

Phase two's own change landed in `08f76fc` and is unchanged. This commit closes
the one thing left in `setup.sh`: the generated MariaDB root password was bound
to a shell variable and then expanded into the shared-infra `.env`. Nothing else
in the file read it, so the variable was one export or one stray `echo` away from
publishing a credential `setup.sh` has no other use for. The `openssl rand`
substitution now runs inside the heredoc, so the value never has a name. The key
stays, because `benchpress/config/docker-compose.yml:8` reads
`${MARIADB_ROOT_PASSWORD}` from that file.

### How to verify

```bash
grep -nE '\$\{?MARIADB_ROOT_PASSWORD' setup.sh    # 0 hits — no expansion of the value
grep -n  'MARIADB_ROOT_PASSWORD'      setup.sh    # 1 hit  — the .env key compose reads
bash -n setup.sh
```

The generated file is byte-identical in shape: `MARIADB_ROOT_PASSWORD=` plus 32
hex characters, same as before.

### Re-verified against the live system

- Four scoped modules, exit code read directly, all `0`: `test_docker_manager`
  (46), `test_mariadb_manager` (22), `test_vpn_adapter` (17),
  `test_deploy_manager` (30). `uvx pre-commit@4.3.0 run --all-files` exit 0, no
  rewrites.
- The phase-two deploy capture re-read independently, not taken on trust:
  68 `exec_create` events as the control, the bench name hit 10 times as a second
  control that the capture works, and 0 hits for the bench's SSH password (len
  16), its code-server password (len 22) and its WireGuard private key (len 44) —
  each length-checked first, because an empty variable makes `grep -cF` match
  every line.
- The base64 payloads decoded: 8 of them, `IDENTIFIED BY` 0, `IDENTIFIED VIA
  mysql_native_password USING` 2 (Docker emits `exec_create` and `exec_start` for
  the same command). The deployed site's own `db_password` appears 0 times raw
  and 0 times decoded.
- The bench still works, which is what the hash change risks: 6 benches
  `Running`, `https://<bench>.benchpress.cloud/api/method/frappe.ping` → 200
  `{"message":"pong"}`, and `frappe.db.count("DocType")` → 318 through the site's
  own database user.
- `docker exec benchpress-mariadb healthcheck.sh --connect --innodb_initialized`
  → exit 0, which is the condition that replaced the credential in the readiness
  loop.

### Still a human's step

`DB_ROOT_PASSWORD` is in the `benchpress_db` healthcheck command and has already
leaked. Rotating a live database password takes every bench off its database
without a coordinated redeploy, so it is not this branch's.

Venkateshvenki404224:
Phase two re-verified on `9568618`. No code changed — this is a correction to the evidence the two comments above cite, and a fresh capture.

### The earlier "0 hits" measurements were invalid

`ssh_password`, `code_server_password` and `admin_password` are Frappe **`Password` fieldtype**. `frappe.db.get_value` returns `"*" * len(value)` for those fields; the real value lives in `__Auth`. Both earlier checks read them that way and grepped the event capture for a string of asterisks.

A length guard does not catch it — the mask keeps the length, so the three measure 16, 22 and 14, exactly as expected. Read them with `frappe.utils.password.get_decrypted_password` instead.

### Re-measured, with the decrypted values

Redeployed `01a5ba36…` through `frappe.enqueue("benchpress.deploy_manager.deploy_bench", queue="long")`; Deploying to Running in 26 s. `docker events --format '{{.Action}}'` captured 262 lines.

| Check | Result |
|---|---|
| `exec_create` — the control | 85 |
| Bench name — second control, a string that is on the stream | 2 |
| SSH password, decrypted (16) | **0** |
| Code-server password, decrypted (22) | **0** |
| Admin password, decrypted (14) | **0** |
| WireGuard private key from `wg0.conf` (44) | **0** |
| Site `db_password` (16), raw and base64-decoded | **0 / 0** |
| base64 payloads decoded — the control for the decode | 6 |
| `IDENTIFIED BY` in decoded SQL | **0** |
| `IDENTIFIED VIA mysql_native_password USING` | 1 |

The conclusion is unchanged. The evidence for it is now real.

### The bench works

- `frappe.db.count("DocType")` returns 318 inside the bench, through the site's own database user — the user created from `SHA1(SHA1(password))`, authenticating with the plaintext.
- SSH accepts the stored password and denies a wrong one. With the masked value it denied both, which is how the fieldtype bug surfaced.
- Site returns 200 publicly and over the tunnel; `wg show` reports the peer up.
- `docker exec benchpress-mariadb healthcheck.sh --connect --innodb_initialized` exits 0.

### How to verify

```bash
```

Four scoped modules pass (46/22/17/30), `uvx pre-commit@4.3.0 run --all-files` exits 0 with no rewrites.

PR #208, opened 2026-08-27
Implemented phase two

Venkateshvenki404224:
## What this changes

`_compose_cmd` runs `benchpress/config/docker-compose.yml` from inside the backend container, so
both `./*.conf` binds reached the host daemon as container paths the host does not have. The
daemon created directories at those paths and both shared services have run on stock defaults
ever since — Redis on `maxmemory 0`, `noeviction` and no memory limit.

Both services now take their settings as `command:` flags:

```yaml
mariadb:  mariadbd --max-connections=500 --innodb-buffer-pool-size=128M --key-buffer-size=16M
redis:    redis-server --maxmemory 256mb --maxmemory-policy allkeys-lru
```

Redis also gains the `mem_limit: 512m` it never had. The buffer pool is pinned at the stock
128M, not the declared 512M: 512M does not fit under the 1g limit (a scratch container reached
779.6 MiB anon and went into swap) and buys nothing at a measured 99.94% hit rate.

Three dead surfaces go with the file nothing read: `DEFAULT_MARIADB_CONFIG`,
`_write_mariadb_config`, and `Database Server.custom_config`. The two config files themselves are
deleted, and `README.md` / `DESIGN_BRIEF.md` no longer name them.

`benchpress/tests/test_shared_compose.py` bans the fault class rather than the instance: no
service in that file may declare a relative bind, in either the short `src:dst` or the long
`{type, source}` form.

## How to verify

```bash
docker compose exec backend bench --site frontend run-tests --module benchpress.tests.test_shared_compose
docker compose exec backend bench --site frontend run-tests --module benchpress.tests.test_mariadb_manager
```

The flags themselves were proved on a scratch pair, which is what keeps this change off the live
containers:

```bash
docker run -d --name p1probe-db --memory 1g -e MARIADB_ROOT_PASSWORD=probe mariadb:10.6 \
  mariadbd --max-connections=500 --innodb-buffer-pool-size=128M --key-buffer-size=16M
docker exec p1probe-db mariadb -u root -pprobe -e \
  "SHOW GLOBAL VARIABLES WHERE Variable_name IN ('max_connections','innodb_buffer_pool_size','key_buffer_size')"
docker rm -f p1probe-db
```

Read back `max_connections 500`, `innodb_buffer_pool_size 134217728`, `key_buffer_size 16777216`,
and on the Redis probe `maxmemory 268435456` with `maxmemory-policy allkeys-lru`. A daemon that
refuses a flag exits rather than ignoring it, which is the loud failure the bind never had.

## Leftovers for a human

**The live pair still runs on stock defaults, deliberately.** A `command:` change applies only on
`docker compose up -d`, and that recreate briefly drops every running bench's database connection
and clears the shared Redis. Interrupting live tenants is a human's call:

```bash
cd benchpress/config && docker compose up -d
docker exec benchpress-redis redis-cli config get maxmemory-policy   # expect allkeys-lru
```

Two empty root-owned directories the daemon created remain on the host at
`/home/frappe/frappe-bench/apps/benchpress/benchpress/config/{mariadb.cnf,redis.conf}`. They are
outside this checkout, they are the live containers' current bind sources, and removing them
needs root — so they are part of the same recreate step, not of this branch:

```bash
sudo rmdir /home/frappe/frappe-bench/apps/benchpress/benchpress/config/{mariadb.cnf,redis.conf}
```

Phase 2 adds the drift check that tells an operator whether the recreate has happened. It will
report drift until it does, and that is correct.

### DIFF
--- a/README.md
+++ b/README.md
@@ -907,7 +907,7 @@ Device management is backed by the **VPN Peer** DocType in the vpn_management ap
 | Schedule | Function | Description |
 |----------|----------|-------------|
 | Every 1 minute | `benchpress.stats_collector.enqueue_stats_sweep` | Enqueues `collect_bench_stats`, which polls Docker CPU/memory/health for running containers (VPN transfer counters are updated by vpn_management's own `poll_status` job) |
-| Every 5 minutes | `benchpress.mariadb_manager.enqueue_health_check` | Enqueues `scheduled_health_check`: shared MariaDB health, restart if down |
+| Every 5 minutes | `benchpress.mariadb_manager.enqueue_health_check` | Enqueues `scheduled_health_check`: shared MariaDB health, restart if down, and logs any live setting that disagrees with the declared flags |
 | Daily at 2 AM | `benchpress.mariadb_manager.enqueue_backup` | Enqueues `scheduled_backup`: full MariaDB backup with 7-day retention |
 
 Every entry here is an enqueuer, not the work itself. Scheduled jobs land on the `default` queue,
--- a/README.md
+++ b/README.md
@@ -725,9 +725,7 @@ benchpress/
 |   +-- hooks.py                  # App config: routes, scheduler, ignore_links_on_delete
 |   +-- mariadb_manager.py        # Shared MariaDB + Redis lifecycle (docker compose)
 |   +-- config/
-|   |   +-- docker-compose.yml    # Shared infrastructure (MariaDB + Redis)
-|   |   +-- mariadb.cnf           # MariaDB custom configuration
-|   |   +-- redis.conf            # Redis custom configuration
+|   |   +-- docker-compose.yml    # Shared infrastructure (MariaDB + Redis), tuned by command flags
 |   |   +-- .env.example          # Environment variable template
 |   |   +-- benchpress-infra.service  # Systemd unit for auto-start on boot
 |   +-- lab-templates/
--- a/benchpress/benchpress/doctype/database_server/database_server.json
+++ b/benchpress/benchpress/doctype/database_server/database_server.json
@@ -21,8 +21,6 @@
   "memory_limit",
   "volume_name",
   "created_at",
-  "configuration_section",
-  "custom_config",
   "error_section",
   "error_message"
  ],
@@ -121,19 +119,6 @@
    "fieldtype": "Datetime",
    "label": "Created At",
    "read_only": 1
-  },
-  {
-   "collapsible": 1,
-   "fieldname": "configuration_section",
-   "fieldtype": "Section Break",
-   "label": "MariaDB Configuration"
-  },
-  {
-   "description": "Full MariaDB server config (INI). Seeded with sensible defaults on creation; edit to tune the server (e.g. max_connections, buffer pool). Re-run setup to apply.",
-   "fieldname": "custom_config",
-   "fieldtype": "Code",
-   "label": "Custom Config",
-   "options": "INI"
   },
   {
    "collapsible": 1,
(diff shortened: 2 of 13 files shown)


Venkateshvenki404224:
## Phase 2 — the platform notices when the settings are wrong

Phase 1 gave both shared services their settings as `command:` flags. Those flags
only take effect when a human recreates the containers, and until now nothing told
an operator whether that had happened. A static test cannot see a flag the daemon
refused, or a manual `SET GLOBAL`, either.

### What this adds

- **`mariadb_manager`** — `mariadb_drift` reads `SHOW GLOBAL VARIABLES` and the
  buffer-pool counters in one round trip; `redis_drift` reads `redis-cli CONFIG GET`.
  Both compare against `DECLARED_MARIADB` / `DECLARED_REDIS`, which name the same
  values `config/docker-compose.yml` declares.
- **`scheduled_health_check`** — the existing 5-minute job now logs every live
  setting that disagrees, and returns the same lines so `bench execute` shows them.
  Drift is logged, never raised: this runs on the scheduler, where an exception is a
  stack trace nobody reads and would take the database check down with it.
- **Logged only when it changes.** Drift stays true until the recreate, so an
  unconditional log would write 288 copies of one row a day into the Error Log it
  exists to be read in. The last reported set goes in the cache; recovery resets it.
- **`diagnostics`** — the MariaDB and Redis rows carry the same comparison, so the
  page an operator already opens stops calling a stock Redis healthy. Drift reads as
  a **Warning**, not an Error: the pair still serves every bench, and it stays
  drifted until a human runs the recreate.
- The MariaDB row always reports the buffer-pool hit rate. The 128M decision rests
  on a single measurement, and the next person should revisit it on evidence.

### How to verify

```bash
docker compose exec backend bench --site frontend run-tests --module benchpress.tests.test_mariadb_manager   # 30 tests
docker compose exec backend bench --site frontend run-tests --module benchpress.tests.test_diagnostics        # 24 tests
docker compose exec backend bench --site frontend execute benchpress.mariadb_manager.scheduled_health_check
docker compose exec backend bench --site frontend execute benchpress.api.run_diagnostics
```

Run against this host, the check reports the drift that is genuinely there:

```
["MariaDB max_connections is 151, declared 500",
 "MariaDB key_buffer_size is 134217728, declared 16777216",
 "Redis maxmemory is 0, declared 268435456",
 "Redis maxmemory-policy is noeviction, declared allkeys-lru"]
```

**That is the correct result, not a failure.** The live pair has not been recreated,
so it still holds stock defaults. A clean result here would mean the comparison is
not reading the live server. `innodb_buffer_pool_size` is absent because the stock
value already equals the declared 128M.

Measured on the live pair:

- The extra read costs **3 queries per run**, against a 0.1 questions/sec baseline.
- The live buffer-pool hit rate reads **99.92%**, independently confirming the 99.94%
  the spec measured — the 512M pool this branch removed buys nothing.
- From a cleared cache key, one run added one Error Log row and the next two added none.
- Both shared containers are still on their original start times
  (`2026-08-24T17:01:47Z`); nothing in this branch touched them.

The Overview page's Shared infrastructure card renders both rows as amber Warnings
with the full comparison and the fix command.

### Leftover — for a human, both of them

1. `cd apps/benchpress/benchpress/config && docker compose up -d` to recreate both
   shared containers. This briefly drops every running bench's database connection
   and clears the shared Redis, which is why an unattended loop does not choose it.
   After it, the check above reports no drift and both rows go green.
2. `sudo rmdir /home/frappe/frappe-bench/apps/benchpress/benchpress/config/{mariadb.cnf,redis.conf}`
   — the two empty root-owned directories the daemon created at the old bind paths.
   Do this after the recreate, since they are the live pair's current bind sources.

PR #209, opened 2026-08-27
Implemented phase three

Venkateshvenki404224:
Pass 1 of the ingress-and-addressing refactor, phase 1 of 3.

## What this adds

`benchpress/addressing.py` is the one place that knows a bench port. It holds
`SITE_HTTP_PORT` and `IDE_HTTP_PORT` and builds every address a bench answers
on. `deploy_manager` and `golden` read their port from it. `api.get_benches`
and `lab_detail._caller_bench` carry the set as `bench["addresses"]`.
`labActions.js` reads that payload and declares no port of its own.

`get_benches` did not select `public_url`. The field is added, or every public
site address returns `None` while the tests stay green.

## What did not move

No dotted string moved, so this phase runs no `migrate`. The route files, the
wildcard anchor, the TLS probe and the `*/5` reconciler stay in
`deploy_manager`. `git grep 'benchpress\.deploy_manager\.'` returns the same
lines it returned before, at shifted line numbers only.

## How to verify

```
bench --site frontend run-tests --module benchpress.tests.test_addressing   # 10 tests
bench --site frontend run-tests --module benchpress.tests.test_deploy_manager
bench --site frontend run-tests --module benchpress.tests.test_api
cd frontend && npx vitest run                                               # 184 tests
bench --site frontend execute benchpress.api.get_benches | grep -c tunnel_site
```

The payload check returns 10 for 10 benches, each with a non-null
`public_site`.

Checked in the browser on a running bench. Lab detail shows the public URL,
the private VPN URL and the code-server address. The primary button opens
`https://<id>.benchpress.cloud` and the IDE button opens
`https://ide-<id>.benchpress.cloud`.


Venkateshvenki404224:
## Phase two — the front door moves out

`benchpress/ingress.py` (new, 256 lines) now owns every write to the Traefik route directory, and it owns the only knowledge of where that directory is. `deploy_manager.py` drops from 1251 to 1029 lines and keeps only the job functions that a dotted string names.

**Four renames**, because `deploy_manager` now calls these across a module boundary and a private name called from another module is not private:

| was | is |
|---|---|
| `_write_instance_route` | `ingress.publish` |
| `_delete_instance_route` | `ingress.withdraw` |
| `_ensure_wildcard_anchor` | `ingress.ensure_anchor` |
| `_log_certificate_state` | `ingress.log_certificate_state` |

**Two new reads.** `published()` and `protected_present()` let `reconcile_instance_routes` stop globbing the directory, so `TRAEFIK_DYNAMIC_DIR` and `PROTECTED_ROUTE_FILES` appear in one file only — a reconcile loop that learned the path and the protected names could take the control plane off the internet through one off-by-one.

`publish` also reads `addressing.IDE_HTTP_PORT` instead of the hardcoded `8080` that phase one left behind.

**No dotted string moves in this phase.** All four job strings still name `benchpress.deploy_manager.*` and all three RQ job ids (`route_sync:<bench>`, `route_reconcile`, `reconcile_instance_routes`) are unchanged. There is no re-export shim — callers reach the new module as `ingress.publish(...)`, never as a from-import, so there is one canonical patch target per function.

### How to verify

```bash
# 1. Tests — read the exit code, not the last line
docker compose exec backend bench --site frontend run-tests --module benchpress.tests.test_image_cache; echo "exit=$?"
docker compose exec backend bench --site frontend run-tests --module benchpress.tests.test_admission; echo "exit=$?"

# 2. The real path writes real files. NOTE: queue-long, not backend — the route
#    directory is bind-mounted read-write into queue-long alone.
docker compose exec queue-long bench --site frontend execute \
  benchpress.deploy_manager.reconcile_instance_routes

# 3. The discriminator for the whole phase — must return nothing
git grep -n '_atomic_write\|TRAEFIK_DYNAMIC_DIR\|PROTECTED_ROUTE_FILES' benchpress/ \
  | grep -v 'benchpress/ingress.py' | grep -v '^benchpress/tests/'
```

### Results on the live deployment

- `test_ingress` **26 tests, exit=0** (four classes moved out of `test_deploy_manager.py`, plus a new `TestDirectoryReads` covering the two reads and the empty-directory case).
- `test_deploy_manager` 88+2, `test_api` 58, `test_image_cache` 11+11, `test_admission` 46 — **all exit=0**. `test_lease` is fenced locally and passes here in the Server check.
- Reconcile returned `{"anchored": false, "written": 6, "deleted": 0, "kept": 2, "attached": {}}`. **`kept: 2`** is the assertion that matters: a 0 or 1 would mean `protected_present` is wrong and a protected file is one pass away from deletion.
- All 8 files in the route directory are **byte-for-byte identical** to a pre-run snapshot, with mtimes unchanged and no temp file left — `_atomic_write`'s content compare still suppressing a pointless Traefik reload. `dynamic.yml` and `wildcard-anchor.yml` untouched; `traefik` never restarted.
- `deleted: 0` with both protected files surviving is the live proof of `published()`: a version that failed to exclude them would have handed them straight to `withdraw`.
- `queue-long` ran four real `sync_instance_route` jobs on the new code, all OK — that is the RQ path where a re-export shim would have raised `ImportError`.
- Imports are clean in **both** orders (`ingress` first and `deploy_manager` first), and `frappe.get_attr` resolves all four job strings.

### One correction to the docs

The route directory is mounted **read-write in `queue-long` and read-only in `traefik`, and nowhere else** — `backend` has no mount. A reconcile run with `docker compose exec backend` raises `TraefikRouteDirectoryMissing` before it writes anything, which is the guard working as designed. Use `queue-long`.

Venkateshvenki404224:
Phase three — the four job strings move.

`sync_instance_route`, `enqueue_route_sync`, `enqueue_route_reconcile`, `reconcile_instance_routes` (now `reconcile`) and `_routable_instance_ips` move from `deploy_manager` to `ingress`. `deploy_manager.py` is 914 lines and holds no routing.

Frappe resolves a scheduled job and an enqueued job from a dotted string at run time, so all four strings are rewritten in the same commit as the move: the `*/5` entry in `hooks.py`, the method in `benchpress_settings.py`, and the two enqueue targets. The operator command in the docstring moves with them. The three RQ job ids are unchanged — a job id is a dedup key, not an import path, so `route_sync:<bench>`, `route_reconcile` and `reconcile_instance_routes` all keep their exact values.

`reconcile` imports `_reconcile_bridge_attachments` inside the function, because `deploy_manager` imports this module. No re-export shim was added: it is green in CI and raises `ImportError` on the path an RQ worker takes.

**Two new static gates in `test_scheduler_entries.py`.** One resolves every method in `scheduler_events`; the other walks every non-test module for literal `frappe.enqueue` targets and resolves those. The walker finds 15 targets today. Both were written and run green on the unmodified tree first, so a pass is evidence the walker works rather than evidence it matched nothing. A later move of a job target now fails a test instead of going quiet.

## How to verify

No string survives:

```
git grep -n 'benchpress\.deploy_manager\.\(sync_instance_route\|reconcile_instance_routes\|enqueue_route_reconcile\|enqueue_route_sync\)'
```

The database agrees, which is the half no static check sees — every `Scheduled Job Type` row resolves under `frappe.get_attr`, no row names the old path, and one names `benchpress.ingress.enqueue_route_reconcile`.

The reconcile converges through its new name. Run it in `queue-long`, not `backend` — `queue-long` is the only container that mounts the route directory:

```
docker compose exec -T queue-long bench --site frontend execute benchpress.ingress.reconcile
```

## Verified live

- The `*/5` cron fired from its new path and the work actually ran: `queue-long` logged `job_name='benchpress.ingress.reconcile' ... Job OK (frontend||route_reconcile)`, under the unchanged job id.
- Live reconcile returned `{"anchored": false, "written": 6, "deleted": 1, "kept": 2, "attached": {}}`. `kept: 2` is both protected files. The one deletion was an orphan route file left by a test bench, which is the reap half working through the new module.
- The six real bench route files are byte-identical, mtimes unchanged. `dynamic.yml` and `wildcard-anchor.yml` untouched. `traefik` never restarted.
- Scoped tests all exit 0: `test_ingress` (50), `test_deploy_manager` (66), `test_scheduler_entries` (5), `test_api` (58), `test_admission` (46).

Note on the phase spec: step 6 suggested `grep -c 'benchpress.ingress' logs/scheduler.log`, expecting a non-zero count. That heuristic is inverted on this site — every line in `scheduler.log` is an `ERROR`, and the old path appeared there 42 times only as `Skipped queueing ... because it was found in queue`. A count of 0 is the healthy state. The worker log above is the real proof.

PR #226, opened 2026-08-27
Implemented phase two

Venkateshvenki404224:
Pass 2 of the five-pass refactor (`todo.md` item 17), phase 1 of two.

## What moved

Seven functions from `docker_manager.py` into `placement.py`:
`subnet_base`, `bench_network_spec`, `bench_network_index`,
`ensure_bench_network`, `ensure_bench_network_for`, `attach_infrastructure`,
`missing_infrastructure` — with `DEFAULT_SUBNET_BASE`, `NETWORK_PREFIX`,
`BRIDGE_DEVICE_PREFIX` and `INFRASTRUCTURE_CONTAINERS`, which only they read.

`ensure_network` and `LEGACY_NETWORK` stay in `docker_manager`: that is the
app's own shared `benchpress` network, with no index, no slot and no headroom.

`missing_infrastructure` stays separate from `attach_infrastructure`. The
reconcile loop needs the read without the write.

Callers reach the moved names as `placement.<name>` — one patch target each.
`diagnostics.py` needed no change, which was the point.

Phase 2 moves the repair half (`_reconcile_bridge_attachments`,
`_record_bridge_network`) out of `deploy_manager.py`.

## How to verify

The bridge functions live in one file:

```
git grep -n 'def bench_network_spec\|def attach_infrastructure\|def ensure_bench_network' benchpress/ \
  | grep -v '^benchpress/placement.py' | grep -v '^benchpress/tests/'
```

returns nothing.

Scoped tests, read `$?` rather than the last line:

```
docker compose exec backend bench --site frontend run-tests --module benchpress.tests.test_placement
docker compose exec backend bench --site frontend run-tests --module benchpress.tests.test_deploy_manager
docker compose exec backend bench --site frontend run-tests --module benchpress.tests.test_golden
docker compose exec backend bench --site frontend run-tests --module benchpress.tests.test_vpn_adapter
```

All five exit 0 (`test_placement` grew from 18 to 29 tests).

Live on staging: a redeploy of a bench on `benchpress-1` completed in 20.7s,
its site answers 200 and `/api/method/ping` answers `pong`, which needs the
database across that bridge. `benchpress-1` now carries `benchpress_traefik`,
`benchpress-mariadb` and `benchpress-redis`. The diagnostics `bridge_capacity`
row still reports: `benchpress-0 6 used / 994 free, benchpress-1 4 used /
996 free — 15990 benches of headroom`.

`placement.py` is 220 lines, under the 300-line rule.


Venkateshvenki404224:
## Phase 2 — the repair pass moves, and pass 1's lazy import goes

`placement` now owns the whole bridge lifecycle: choose, describe, create, attach, report, repair, record. `deploy_manager` has no networking left in it.

**Moved out of `deploy_manager`**

| Was | Is |
|---|---|
| `_reconcile_bridge_attachments()` | `placement.repair()` |
| `_record_bridge_network(bench, network)` | `placement.record_bridge_network(bench, network)` |

Both are called from another module, so neither keeps its underscore. Both docstrings move verbatim — they carry the two facts a reader would otherwise undo: that `docker compose up -d traefik` is what strips the bridges, and that `record_bridge_network` writes the column directly because `Bench Instance.validate` refuses the field to a tenant on a non-`Draft` row.

**The lazy import is gone.** `ingress.reconcile` opened with `from benchpress.deploy_manager import _reconcile_bridge_attachments` and a comment predicting this pass would remove it. It now calls `placement.repair()` through a module-level import. There is no cycle left: `ingress` imports `placement`, and `placement` imports neither. No re-export shim was added to `deploy_manager`.

**`repair()` reports what it found as well as what it fixed.** It returns `{"attached": ..., "missing": ...}` instead of `{"attached": ...}` alone. Previously a bridge that was missing all three containers and could not reattach any of them returned `{"attached": {}}` — byte-identical to a healthy tick. Item 8's reconcile loop needs those apart.

`ingress.reconcile()`'s dict gains the `missing` key with it.

**Tests.** `TestReconcileBridgeAttachments` moves to `test_placement.py` as `TestRepair`, plain `unittest.TestCase` to match that file, reusing its `_network_holding` helper. One test is new: `test_what_it_found_is_reported_apart_from_what_it_fixed`.

### How to verify

```bash
docker compose exec backend bench --site frontend run-tests --module benchpress.tests.test_placement; echo "exit=$?"

# the lazy import is gone, and no cycle came back from either entry point
git grep -n 'from benchpress.deploy_manager import' benchpress/ingress.py     # empty
docker compose exec backend env/bin/python -c "import benchpress.placement, benchpress.ingress"
docker compose exec backend env/bin/python -c "import benchpress.ingress, benchpress.placement"

# the route directory is bind-mounted read-write only in queue-long
docker compose exec queue-long bench --site frontend execute benchpress.ingress.reconcile
```

To prove the repair still repairs, detach an infrastructure container from a bridge carrying **no running bench** and run the pass again:

```bash
docker network disconnect benchpress-2 benchpress_traefik
docker compose exec queue-long bench --site frontend execute benchpress.ingress.reconcile
docker network inspect benchpress-2 --format '{{range .Containers}}{{.Name}} {{end}}'
```

Done on an empty bridge here: the read reported `{"benchpress-2": ["benchpress_traefik"]}`, `repair()` returned that under both `missing` and `attached`, and Docker confirmed Traefik back on the bridge. `run_diagnostics` still reports `bridge_capacity` as `pass`.

Venkateshvenki404224:
Follow-up commit `343e519` — `fix(ingress): report the route half where the directory is not mounted`.

Phase 2 verification step 4 runs the convergence pass from `backend`:

```
docker compose exec backend bench --site frontend execute benchpress.ingress.reconcile
```

That raised. `reconcile` did the bridge repair, then went into `ensure_anchor`, which writes through `_atomic_write` and refuses a container with no route directory. Only `queue-long` mounts `config/traefik/generated/dynamic`. `bench execute` wraps the lookup and the call in one `try`, so the traceback surfaced as `NameError: name 'benchpress' is not defined` and named neither the mount nor the queue.

`reconcile` now checks the mount before it writes, and reports the route half as `None`. Zero counts would read as a converged pass. `None` says the directory was never opened. The writers still raise — that contract is unchanged, and its three tests still pass.

The cron cannot reach the new branch: it goes through `enqueue_route_reconcile`, which pins `queue="long"`.

## Verify

```
# backend — no mount, reports instead of raising
docker compose exec backend bench --site frontend execute benchpress.ingress.reconcile
# {"anchored": null, "written": null, "deleted": null, "kept": null, "attached": {}, "missing": {}}

# queue-long — the real pass, unchanged
docker compose exec queue-long bench --site frontend execute benchpress.ingress.reconcile
# {"anchored": false, "written": 6, "deleted": 0, "kept": 2, "attached": {}, "missing": {}}
```

`kept: 2` is `dynamic.yml` and `wildcard-anchor.yml`, both untouched.

Repair still repairs, checked against the live daemon on the empty bridge `benchpress-2`: detaching `benchpress_traefik` made `placement.repair()` return `{"attached": {"benchpress-2": ["benchpress_traefik"]}, "missing": {"benchpress-2": ["benchpress_traefik"]}}`, and `docker network inspect` shows it back.

`test_ingress` 23 + 28 tests exit 0, and `test_placement` / `test_deploy_manager` / `test_docker_manager` / `test_golden` all exit 0. `uvx pre-commit@4.3.0 run --all-files` passes all 12 hooks.

PR #228, opened 2026-08-27
Implemented phase two

Venkateshvenki404224:
Pass 3 of the five-pass refactor (todo item 14), phase 1 of 2.

## What this adds

`benchpress/site_names.py` owns the site name end to end: `qualify`, `claim`, `claimed`, `release`, and the two DNS constants. It imports neither `api` nor `deploy_manager`, so the import cycle those two dodged is gone rather than worked around — no re-export shim anywhere.

- `docker_manager.resolve_site_name` -> `site_names.qualify`
- `api._claim_site_name` -> `site_names.claim` (public: three modules call it)
- `deploy_manager._claimed_site` -> `site_names.claimed`, and the function-scope `from benchpress.api import _claim_site_name` is deleted
- `release` is new as a function. `BenchInstance.on_trash` deleted the `Bench Site` rows inline; naming it puts all four operations in one place.

`deploy_manager.py` drops from 874 to 856 lines. No schema change, no `migrate`, no dotted string moved.

## How to verify

```bash
docker compose exec backend bench --site frontend run-tests --module benchpress.tests.test_site_names
docker compose exec backend bench --site frontend run-tests --module benchpress.tests.test_api
docker compose exec backend bench --site frontend run-tests --module benchpress.tests.test_deploy_manager
```

Read `$?`, not the last line — these targets print two summaries.

The cycle, from every entry point (all four must succeed):

```bash
git grep -n 'from benchpress.api import' benchpress/deploy_manager.py   # must be empty
docker compose exec -w /home/frappe/frappe-bench/sites backend ../env/bin/python -c "import benchpress.site_names, benchpress.api, benchpress.deploy_manager"
docker compose exec -w /home/frappe/frappe-bench/sites backend ../env/bin/python -c "import benchpress.deploy_manager, benchpress.api"
docker compose exec -w /home/frappe/frappe-bench/sites backend ../env/bin/python -c "import benchpress.api, benchpress.deploy_manager"
```

A cycle that breaks from only one entry point passes CI and fails in every RQ worker, which is why all orders are checked.

## Tests

`test_site_claim.py` becomes `test_site_names.py` and absorbs `TestResolveSiteName` from `test_docker_manager.py`. New: a second `claim` on one name is refused and names the owning bench (the contention case the suite never had), a same-bench reclaim is quiet, and `release` frees every name a bench holds.

## Not in this phase

`full_domain` is untouched and still drifts, and teardown still drops both candidate names. Phase 2 ends that.


Venkateshvenki404224:
## Phase 2 — `full_domain` goes, and teardown drops one name

`Bench Site.full_domain` was recomputed from the free-text `Bench Instance.domain` on every save, so it drifted from the name the site's database was created under. `get_database_name` hashes that name, so teardown dropped a hash that never existed and orphaned the real database. The workaround was to drop both candidate names.

Nothing resolved the field — the site is served by its container on its WireGuard address — so it is deleted end to end rather than moved. `_drop_bench_site_databases` now reads `site_name` and drops exactly one name.

**Gone from:** the DocType JSON, the controller (`compute_full_domain` and its `validate` hook), `api.py`, `overview.py`, `lab_detail.py`, `labActions.js`, `SitesCard.vue`, and the four tracked `.md` files that listed it. `git grep -n full_domain -- ':!benchpress/public'` returns nothing.

**Beyond the spec's letter:** `bench migrate` removed the DocField but left the column, so `benchpress.patches.drop_bench_site_full_domain` drops it, following the existing `drop_burn_fields` pattern. The spec only asked that a leftover column be reported; the repo's convention is to drop it, so it is dropped. `DESCRIBE` now has no `full_domain` row.

### How to verify

```bash
docker compose exec backend bench --site frontend run-tests --module benchpress.tests.test_api; echo "exit=$?"
docker compose exec backend bench --site frontend run-tests --module benchpress.tests.test_site_names; echo "exit=$?"
cd frontend && yarn test:run
git grep -n full_domain -- ':!benchpress/public'   # must be empty
```

`test_teardown_drops_only_the_name_the_site_was_created_under` asserts one name is dropped **and** that a drifted second name is not. That negative control was confirmed to have teeth: re-introducing a second candidate name makes it fail.

### Verified live on staging

- A throwaway CRM bench deployed as `snp2drift.benchpress.cloud`, its `domain` then set to `drifted.example` so the old code would have drifted.
- Real database `_391f55f0d72dc430`; the old code would have also dropped `_0c7d0483a4704eab`, a hash that never existed.
- After delete: real database gone, `_`-prefixed database count 15 → **14** — exactly one dropped, no collateral. Container removed, `Bench Site` row released.
- Lab detail Sites card reads `snp2drift.benchpress.cloud` (`Active`, `Open` enabled), so `siteLabel` losing the head of its fallback chain shows the site name rather than a blank.

Venkateshvenki404224:
### Correction to the note above: the column patch was reverted (`2ac2b1a`)

I added `benchpress.patches.drop_bench_site_full_domain` to drop the column `bench migrate` leaves behind, following the `drop_burn_fields` pattern. It had to name `full_domain` in order to delete it, which put the field back into tracked source and contradicted this phase's own gate — the field must survive in no tracked file. So it is reverted.

What that means in practice:

- `bench migrate` removes the **DocField**, so nothing can read or write the value. `frappe.get_meta("Bench Site").get_field("full_domain")` is `None`.
- A **fresh** site never gets the column, because the DocType no longer declares it.
- An **existing** site keeps an inert column that no field and no code path can reach. Harmless, and the reason the field could be deleted at all: it was derived, so no value is lost.

If dropping that column is wanted later, it is a one-file patch — but it cannot coexist with the "no tracked file names `full_domain`" rule as written.

PR #231, opened 2026-08-27
Implemented phase two

Venkateshvenki404224:
Phase 1 of the `fake-docker-test-seam` spec: one stateful in-memory Docker client, installed at `docker_manager.get_client`, with the real modules running against it.

## What this adds

- `benchpress/tests/fakes.py` — `FakeDocker` (containers, networks, volumes, images, `info`, and the nine container methods) and `FakeDockerMixin`.
- `benchpress/tests/test_fakes.py` — the fake has its own tests, or a bug in it reads as a bug in the app. `containers.list` is tested in both directions.
- `TestDockerManagerBlockIO` and `TestDeployManager` take the mixin. `TestDeployManager` drops from 42 patch calls to 39; phase 2 takes it to 13.

No production file is touched.

## Two things worth knowing

`get_client` is bound at import in `mariadb_manager` and `diagnostics`, so patching `docker_manager` alone misses them. `TestGetClientImporters` re-derives that list from source with `ast`, so a new module-level import fails a test rather than silently reaching a real daemon.

`DOCKER_HOST` is not a kill switch for this app: `get_client` passes an explicit `base_url` from `BenchPress Settings` and never reads the environment. The conversion was verified by refusing `docker.DockerClient.__init__` for a whole run instead.

## How to verify

```bash
docker compose exec backend bench --site frontend run-tests --module benchpress.tests.test_fakes; echo "exit=$?"
docker compose exec backend bench --site frontend run-tests --module benchpress.tests.test_docker_manager; echo "exit=$?"
```

To prove the fake is installed rather than trusted, put a `sitecustomize.py` on the path that raises from `docker.DockerClient.__init__` and run the same three modules. All three stay green.

That run also surfaced a finding outside this phase's scope: 25 tests in `TestDeployStepMarkers` and `TestTerminalStateNotifications` do reach a real daemon today. They are not part of `TestDeployManager`, so the spec leaves them; the mixin is available whenever someone next touches them.

### DIFF
--- a/benchpress/tests/fakes.py
+++ b/benchpress/tests/fakes.py
@@ -1,0 +1,283 @@
+# Copyright (c) 2026, Venkatesh and Contributors
+# See license.txt
+
+"""One stateful in-memory Docker, installed at `docker_manager.get_client`.
+
+Tests read state (`self.docker.created[-1]`) rather than mock call records.
+"""
+
+from unittest.mock import patch
+
+import docker
+
+from benchpress import diagnostics, docker_manager, mariadb_manager
+
+# The modules that bind `get_client` at import instead of reaching it through
+# `docker_manager`, so `patch.object(docker_manager, ...)` alone misses them.
+# `test_fakes.TestGetClientImporters` re-derives this from source: a new
+# module-level import fails a test rather than reaching a real daemon.
+GET_CLIENT_MODULES = (docker_manager, mariadb_manager, diagnostics)
+
+DEFAULT_BLOCK_DEVICES = ["/dev/sda"]
+DEFAULT_INFO = {"Runtimes": {"runc": {}, "sysbox-runc": {}}, "DefaultRuntime": "runc"}
+DEFAULT_STATS = {
+	"cpu_stats": {
+		"cpu_usage": {"total_usage": 200_000_000},
+		"system_cpu_usage": 4_000_000_000,
+		"online_cpus": 2,
+	},
+	"precpu_stats": {"cpu_usage": {"total_usage": 100_000_000}, "system_cpu_usage": 2_000_000_000},
+	"memory_stats": {"usage": 128 * 1024 * 1024, "limit": 512 * 1024 * 1024},
+}
+
+
+class UnscriptedExec(AssertionError):
+	"""A strict FakeDocker was asked to run a command no test registered."""
+
+
+class FakeContainer:
+	def __init__(self, client, container_id, name, labels, network, runtime, ip, status="created"):
+		self.client = client
+		self.id = container_id
+		self.name = name
+		self.labels = labels or {}
+		self.status = status
+		self.health = ""
+		self.stats_response = dict(DEFAULT_STATS)
+		self.archives: dict[str, bytes] = {}
+		self.attrs = {
+			"HostConfig": {"Runtime": runtime or "runc", "NetworkMode": network},
+			"NetworkSettings": {"Networks": {network: {"IPAddress": ip}}, "IPAddress": ip},
+			"State": {"Health": {"Status": self.health}},
+		}
+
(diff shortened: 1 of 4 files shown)


Venkateshvenki404224:
## Phase 2 — `TestDeployManager` converted, 39 patch calls to 13

`TestDeployManager` now runs the real `docker_manager`, `mariadb_manager` and `vpn_adapter` code against the in-memory Docker client from phase 1. Same 17 tests, same assertions, 26 fewer patches. No production file changed across either phase.

### What the 26 removed patches became

Assertions that read a mock's call record now read the fake's state:

| was | is |
|---|---|
| `mock_stop.assert_called_once_with("container-stop-test")` | `self.docker.stopped == ["container-stop-test"]` |
| `mock_remove_container.assert_called_once_with(...)` | `self.docker.removed == [bench.bench_name]` |
| `mock_drop_db.assert_called_once_with(server, site)` | the `DROP DATABASE` the db container's exec actually carried |
| `mock_remove_peer.assert_called_once()` | the `Cleanup: VPN peer removed` line the run wrote |

The two deploy-failure tests drive their failure through the fake instead of a mock side effect. The daemon refuses the container start in one and `docker compose up redis` fails in the other, so both run the real rollback path end to end rather than a patched stand-in for it.

### The 13 that stay

Exactly the three populations no seam retires:

- **8 `deploy_manager` self-patches** — six `_deploy_bench`, two `_notify_owner`. Item 13's problem: splitting the module removes them, a seam cannot.
- **4 Frappe and cache patches** — `frappe.enqueue` ×2, `frappe.msgprint`, `is_job_enqueued`. No injection point, and patching them is idiomatic.
- **1 `ingress.ensure_anchor`** — it writes Traefik's wildcard anchor on the host, which no test may touch.

`_remove_stale_container` looked like a self-patch and was not: its body is pure Docker, so the fake retires it.

### Added to the fake

Only what the conversion needed — per-container exec history, stop and removal records, a container that refuses to start (which is also the only way `start_bench_container`'s address-pool roll becomes testable), and `sql_of` to read back what `execute_sql` base64-encoded. The mixin now also drops the memoised `host_runtimes` read: it lives on `frappe.local`, outlives a test, and would otherwise let a real daemon read survive the fake. Every addition has a case in `test_fakes.py`.

### How to verify

```bash
docker compose exec backend bench --site frontend run-tests --module benchpress.tests.test_fakes; echo "exit=$?"

# 13, and none of them naming an engine module
awk '/^class TestDeployManager\(/,/^class TestDeployStepMarkers\(/' benchpress/tests/test_deploy_manager.py | grep -c 'patch(\|patch\.object('
awk '/^class TestDeployManager\(/,/^class TestDeployStepMarkers\(/' benchpress/tests/test_deploy_manager.py | grep 'patch(\|patch\.object(' | grep -E 'docker_manager|mariadb_manager|vpn_adapter'

# 17 tests, unchanged
awk '/^class TestDeployManager\(/,/^class TestDeployStepMarkers\(/' benchpress/tests/test_deploy_manager.py | grep -c 'def test_'

# no production file touched
git diff --name-only origin/version-16... | grep -v '^benchpress/tests/'
```

Run with the real daemon refused (`docker.DockerClient.__init__` raising, via a `sitecustomize` on `PYTHONPATH`), all 17 tests in the class pass. The 21 failures that run produces are all in `TestDeployStepMarkers` and `TestTerminalStateNotifications` — the classes this phase deliberately left alone, which reach a real daemon today and pass only because CI has one. Pass 5 rewrites them when their code moves to `lifecycle.py`.

### Handed forward

- **Item 13** owns its eight self-patches by name; splitting `deploy_manager` is the only thing that removes them.
- **Item 9** can change `create_bench_container`'s return type and test it without a host that has xfs or four cores.
- **Item 8** inherits a `containers.list` harness that predates `list_benches()`.

PR #233, opened 2026-08-27
Implemented phase three

Venkateshvenki404224:
Phase 1 of the bench-lifecycle-module spec: `benchpress/lifecycle.py` becomes the only writer of `Running`.

- `running(bench, action=...)` carries the four side effects in the order that is the contract: the container call, the status, metering before the save, the route sync after the commit.
- `api._start_bench`, `BenchInstance.enqueue_start` and `api._restart_in_grace` are now wrappers. The duplicate start path is deleted, and the grace restart buys its metering window instead of skipping it.
- `deploy_bench`, `_deploy_bench`, `redeploy_bench` and `_redeploy_bench` move to `lifecycle`; the three dotted job strings that named them are rewritten in the same commit. `deploy_bench:{name}` is unchanged.
- `test_lifecycle.TestOneWriterPerStatus` AST-walks the non-test tree and fails if a fifth writer of `Running` appears.

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -20,7 +20,8 @@ apps/benchpress/
 ├── benchpress/
 │   ├── hooks.py               # App config, scheduler, routes
 │   ├── api.py                 # REST API (~20 endpoints)
-│   ├── deploy_manager.py      # Build & deploy orchestration
+│   ├── lifecycle.py           # Bench transitions: running, and the deploy that reaches it
+│   ├── deploy_manager.py      # Image building, site creation, stop and teardown
 │   ├── docker_manager.py      # Docker SDK wrapper
 │   ├── vpn_adapter.py         # Seam to the vpn_management VPN plane
 │   ├── stats_collector.py     # Container stats cron job
@@ -88,9 +89,9 @@ This is the **brain** of BenchPress. It coordinates builds and deployments.
 | Function | Called By | What It Does |
 |----------|-----------|--------------|
 | `build_lab(lab_name)` | Background job from `build_lab_image` API | Builds Docker image for a Lab. Creates Build Log doc, streams logs via WebSocket (`lab_build_log` event). Sets Lab status to Ready or Error. |
-| `deploy_bench(bench_name)` | Background job from `create_bench` API | **Main deploy pipeline**: check image → remove stale container → create container → start → register VPN Peer + configure container tunnel (via `vpn_adapter`) → set SSH password → mark Running |
+| `lifecycle.deploy_bench(bench_name)` | Background job from `create_bench` API | **Main deploy pipeline**: check image → remove stale container → create container → start → register VPN Peer + configure container tunnel (via `vpn_adapter`) → set SSH password → mark Running |
 | `stop_bench(bench_name)` | Background job from `bench_action` | Stop container |
-| `redeploy_bench(bench_name)` | Background job from `bench_action` | Stop + remove container, reset to Draft, call deploy_bench |
+| `lifecycle.redeploy_bench(bench_name)` | Background job from `bench_action` | Stop + remove container, reset to Draft, call deploy_bench |
 | `log_deploy(bench_name, msg, type)` | Internal helper | Saves Deploy Log + publishes `bench_deploy_log` WebSocket event |
 
 ### `docker_manager.py` (191 lines) — Docker SDK Wrapper
@@ -273,7 +274,7 @@ static markup of a session that is not running anywhere.
                        ▼
 ┌─────────────────────────────────────────────────────────────┐
 │ 3. DEPLOY BENCH                                             │
-│    api.create_bench() → enqueue deploy_manager.deploy_bench │
+│    api.create_bench() → enqueue lifecycle.deploy_bench      │
 │    Status: Deploying                                        │
 │    Pipeline:                                                │
 │      a. Check image exists (build if not)                   │
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -90,7 +90,7 @@ This is the **brain** of BenchPress. It coordinates builds and deployments.
 |----------|-----------|--------------|
 | `build_lab(lab_name)` | Background job from `build_lab_image` API | Builds Docker image for a Lab. Creates Build Log doc, streams logs via WebSocket (`lab_build_log` event). Sets Lab status to Ready or Error. |
 | `lifecycle.deploy_bench(bench_name)` | Background job from `create_bench` API | **Main deploy pipeline**: check image → remove stale container → create container → start → register VPN Peer + configure container tunnel (via `vpn_adapter`) → set SSH password → mark Running |
-| `stop_bench(bench_name)` | Background job from `bench_action` | Stop container |
+| `lifecycle.stopped(bench_name)` | Background job from `bench_action` | Stop container |
 | `lifecycle.redeploy_bench(bench_name)` | Background job from `bench_action` | Stop + remove container, reset to Draft, call deploy_bench |
 | `log_deploy(bench_name, msg, type)` | Internal helper | Saves Deploy Log + publishes `bench_deploy_log` WebSocket event |
 
--- a/benchpress/api.py
+++ b/benchpress/api.py
@@ -6,7 +6,7 @@ from frappe import _
 from frappe.query_builder import DocType
 from frappe.query_builder.functions import Count
 
-from benchpress import addressing, image_cache, lab_detail, lab_templates, labs, site_names
+from benchpress import addressing, image_cache, lab_detail, lab_templates, labs, lifecycle, site_names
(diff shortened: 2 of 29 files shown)


Venkateshvenki404224:
**Phase 1 — one writer for `Running`.** CI green.

`benchpress/lifecycle.py` now holds `running(bench, action=...)` and is the only code in the app that writes `status = "Running"`. `api._start_bench`, `BenchInstance.enqueue_start` and `api._restart_in_grace` are wrappers; the duplicate start path is deleted. `deploy_bench`, `_deploy_bench`, `redeploy_bench` and `_redeploy_bench` moved with it, taking `deploy_manager` from 856 to 604 lines.

The grace restart now buys its metering window by construction — before this it reached `Running` with the deadline that had stopped it, and the next sweep stopped it again.

**How to verify**

```bash
git grep 'benchpress\.deploy_manager\.\(deploy_bench\|redeploy_bench\)'   # empty
bench --site frontend run-tests --module benchpress.tests.test_lifecycle  # read $?
```

`TestOneWriterPerStatus` AST-walks the non-test tree and fails if a fifth writer of `Running` appears, so the deletion cannot be undone quietly.

Driven against a real bench: full deploy (11/11 steps, 182 s), SPA start, SPA restart (`started_at` preserved), Desk `enqueue_start`, and the route file deleted on stop and re-created on start. Pressing Deploy twice refuses the second — `deploy_bench:{name}` is unchanged.

Two notes for review: credits are off on staging, so the grace-restart deadline is proved by tests that arm credits for their own transaction rather than by flipping a live control-plane switch; and `lifecycle` is imported at module scope in `api` and `bench_instance` because a lazy import let a `docker_manager` patch bind itself into the module permanently — the cause of the CI-only failure in the first push.

Venkateshvenki404224:
## Phase 2 — one writer for `Stopped`

`deploy_manager.stop_bench` becomes `lifecycle.stopped`. The five side effects a
stop must not forget now sit in one function: the lease record, the status, the
metering close, the admission release and the `Bench Site` deactivation.
`_deactivate_bench_sites` moves with it; `teardown_bench` reaches it through a
module-scope import until phase 3 takes teardown.

**Two dotted strings moved in the same commit** — `credits/lease.py:309` and
`credits/sweep.py:142`. `git grep 'benchpress\.deploy_manager\.stop_bench'` is
empty. The `stop_bench:{name}` job id did **not** move: it is a dedup key, and
renaming it would let two stops of one bench overlap.

Four other test modules called the function directly and the spec had not counted
them — `test_lease`, `test_lease_drain`, `test_ingress`, `test_site_names`. Their
`patch.object(deploy_manager, "stop_container")` targets move to `lifecycle` too,
except the one wrapping `teardown_bench`, which still binds it from `deploy_manager`.

### How to verify

```bash
git grep 'benchpress\.deploy_manager\.stop_bench'          # empty
bench --site frontend run-tests --module benchpress.tests.test_lifecycle
```

Verified against the live deployment on a bench deployed for the purpose
(`44d0314…`, owner `bl-phase2-tenant@example.com`, lab `crm`) — never anyone else's:

| Check | Result |
|---|---|
| SPA stop (`api.bench_action`) | `Stopped`, site row `Inactive`, route file gone, slots 1 → 0 |
| Desk stop (`Bench Instance.enqueue_stop`) | same end state, slots 1 → 0 |
| Expiry sweep (enqueued dotted string) | a real RQ worker resolved `method='benchpress.lifecycle.stopped'`, `Job OK`, under the unchanged job id `stop_bench\|44d0314…` |
| Admission slot released | `active_instances` 1 → 0 and the `Bench Admission` row deleted, after every path |
| Failure path | `stop_container` raising re-raises, and after an explicit `rollback()` the lease release survives (`expiry_attempts: 1`) — it committed first |

Ordered as the move requires: rewrite every string → `migrate` → restart the
workers. `queue-long` depth was 0 before the restart.

### Tests

The AST walk that proves one writer of `Running` now also proves one writer of
`Stopped`. It asserts `{lifecycle.py, mariadb_manager.py}`: `mariadb_manager`
writes `Stopped` on `Database Server`, a different doctype's status.

`TestStopped` drives the transition itself — status, container, the no-container
case, site deactivation and its idempotence, the admission release, the route sync
ordering, and the failure-path commit. The admission release had no assertion in
this file, and it has no visible symptom until a caller hits their cap.

Venkateshvenki404224:
Phase 2 follow-up — `f690422 refactor(lifecycle): name the Database Server statuses`.

`8ffe656` left one hole. `mariadb_manager` assigned the literal `status = "Stopped"` on a **`Database Server`**, a different doctype from `Bench Instance`. Neither a grep for the one writer of `Stopped` nor the AST guard in `test_lifecycle` can tell the two apart, so the guard carried `mariadb_manager.py` as a named exception and the independent check still reported two writers.

This commit names the four `Database Server` statuses as module constants. No behaviour changes: every literal becomes the constant that holds it. `lifecycle.stopped` is now the only code in the package that assigns the string, and the guard asserts a one-element set.

**How to verify**

```bash
# the check that used to report two writers
git grep -l 'status = "Stopped"' -- 'benchpress/*.py' ':!benchpress/tests/*' | grep -v lifecycle.py    # empty

docker compose exec backend bench --site frontend run-tests --module benchpress.tests.test_lifecycle; echo "exit=$?"
```

Scoped suites all exit 0: `test_lifecycle` (14), `test_mariadb_manager` (31), `test_api` (58), `test_deploy_manager` (58), `test_bench_instance` (19), `test_credit_sweep` (19), `test_stats_collector` (5). `uvx pre-commit@4.3.0 run --all-files` is clean.

Driven against the live site on a bench deployed for the purpose (`44d0314…`, `bl-phase2-tenant@example.com`), start then stop through `api.bench_action`: `Stopped → Running` claims the admission slot, `Running → Stopped` releases it, and the route file is gone afterwards. Every `Scheduled Job Type` row resolves; `benchpress.lifecycle.stopped` and `.running` both resolve under `frappe.get_attr`.

Venkateshvenki404224:
## Phase 3 — teardown, and the failure path that leaked

`deploy_manager.teardown_bench` → `lifecycle.torn_down`, with `_drop_site_database`
beside it. `_cleanup_failed_deploy` becomes `lifecycle.errored`. Both endings run
through one `_ended` helper carrying the route withdrawal, the site rows, the
metering settle and the admission release, so each of the three transitions now
has one writer.

**The leak this closes.** `ingress.publish` and `_record_primary_site` both run
before the deploy steps that can still fail, and `_cleanup_failed_deploy`
withdrew no route and touched no site rows. A deploy that died at the SSH step
left a public hostname answering and a `Bench Site` row reading `Active` for a
bench with no container.

`errored` keeps the site database where `torn_down` drops it — the row is the
name claim, and a retry drops the database itself before recreating the site. It
clears only *this* run's addresses, so a failure before the container leaves the
previous deploy's alone.

### How to verify

```bash
git grep 'teardown_bench'                                  # empty
git grep -l 'def _cleanup_failed_deploy' -- 'benchpress/*.py' ':!benchpress/tests/*'   # empty
git grep -l 'status = "Running"\|status = "Stopped"\|status = "Draft"\|status = "Error"' \
  -- 'benchpress/*.py' ':!benchpress/tests/*' | grep -v lifecycle.py   # empty

bench --site frontend run-tests --module benchpress.tests.test_lifecycle        # 26
bench --site frontend run-tests --module benchpress.tests.test_deploy_manager   # 58
bench --site frontend run-tests --module benchpress.tests.test_api              # 58
```

`test_a_failed_deploy_leaves_no_route_file_answering` and
`..._no_site_row_reading_active` both fail if the route withdrawal or the site
deactivation is removed from `_ended` — checked, not assumed.

### Driven live

Bench `b3bf8b03df4a5d59a965eac2419ce485` was deployed for this purpose (owner
`bl-phase3-tenant@example.com`); no tenant's bench was used.

| Step | Result |
|---|---|
| Deploy → `Running` | route file written, site row `Active` |
| Redeploy failing at `linkuser.sh` (past both) | **0 route files**, site row `Inactive`, name kept |
| Retry | `Running`, same site name, row back to `Active` |
| Teardown via `queue-long` | no container, no route file, database `_dc73a462634c624a` dropped, `Draft` |
| `ingress.reconcile` | `written: 7, deleted: 0, kept: 2` |
| `Scheduled Job Type` rows | all resolve |

No dotted string moved, so no `Scheduled Job Type` row changed; `migrate` was run
anyway and exited 0. Host surfaces byte-identical to preflight, traefik
`restarts=0`.

### Judgment calls recorded

- **`deploy_manager` keeps its name.** What is left is image building and site
  creation, which is still deploying, and renaming it would move
  `api.py`'s `benchpress.deploy_manager.build_lab` string for no gain.
- **`lifecycle.py` is not split** at the spec's 300-line trigger. The size is
  `_deploy_bench`, not the teardown half; moving `torn_down` out would take
  `_deactivate_bench_sites` with it, which `stopped` also uses — a cycle — and
  would contradict "all three transitions live in `lifecycle.py`".
- **Docstrings are two lines** because the repo's `docstring_limit` hook caps
  them; what they carried is now comments on the lines they constrain.

PR #246, opened 2026-08-27
Implemented phase four

Venkateshvenki404224:
Phase 1 of the per-plan-resource-limits spec: the payer is the bench's owner.

The admission hold is the only thing in the app that can refuse a start, and it was priced against the caller rather than the account `metering` debits. Measured on the dev host: an owner at 1.0 credits driven to -4.0 by a caller holding 75.

- `guard._enforce` resolves a payer and prices the claim, the concurrency cap and the balance check against it.
- `requires_admission` takes a `payer` callable; `build_lab_image` names the lab's author.
- `cap_builds_per_day` counts the lab owner's rows. Counted against the caller it could never fire — the endpoint is admin-only.
- A fifth repair rule re-keys claims already on the wrong account, placed before `_heal_counters` so the heal reads corrected rows.
- `test_lease` is unfenced and its entry deleted from the ralph-loop reference.

Verified live against the dev site, credits forced in memory only and everything rolled back:

| Step | Result |
|---|---|
| Start of an owner's bench by a funded caller | `REFUSED — Not enough credits: this needs 5.0 and 1.0 are available — 4.0 short`; owner's balance unmoved, no row claimed on either account |
| Re-key a planted wrong-account claim | row moved to the owner, counters `(1, 5.0)` / `(0, 0.0)`, second pass re-keys nothing |
| Build at the owner's daily cap | `REFUSED`, with the under-cap control admitted |
| `api.get_benches` | 12 benches, every one at its expected status |

Tests, reading the exit code: `test_lease` 68 OK (was 1 error), `test_admission` 53 OK, `test_credits`, `test_api`, `test_credit_guard`, `test_signup`, `test_docker_manager`, `test_ingress` all exit 0. `pre-commit run --all-files` clean.

### DIFF
--- a/.claude/skills/issue-to-phases/references/ralph-loop.md
+++ b/.claude/skills/issue-to-phases/references/ralph-loop.md
@@ -87,6 +87,24 @@ subject — and verify it by hand before trusting it.
 an unguarded `grep` miss aborts the whole loop instead of reporting the failure
 the gate exists to catch. End helpers with `|| true` and an explicit `return 0`.
 
+**A helper whose stdout is the signal never counts with `grep -c`.** It prints
+`0` on no match, so `[ -z "$(helper)" ]` is false for a clean tree and false for
+a dirty one: the gate is unsatisfiable and fails every phase whatever the agent
+writes. Emit the matches themselves — `grep -o`, `grep -l` — and let empty mean
+clean. Guarding the exit code with `|| true` does not help, because the `0` is on
+stdout.
+
+> This one shipped. Two phase-1 gates read `grep -c … || true` and tested the
+> result with `-z`, so a correct implementation was retracted as a failure. Prove
+> every helper both ways before shipping the loop: break the thing on purpose and
+> watch the gate go red, then fix it and watch it go green. A gate only ever
+> observed on a clean tree is untested.
+
+**Beware an `awk` range whose start line also matches its end pattern.**
+`/def _enforce/,/^def [a-z_]+\(/` looks like "the body of `_enforce`" and selects
+the `def` line alone, because that line matches the end pattern too. The body is
+never read, so the gate sees nothing and reports clean.
+
 **Redirect stdin on every `docker compose exec`, or the loop hangs forever.**
 GNU `timeout` calls `setpgid()` to put its child in its own process group so it
 can kill the group. That group is *background* relative to the terminal, so the
--- a/.claude/skills/issue-to-phases/references/ralph-loop.md
+++ b/.claude/skills/issue-to-phases/references/ralph-loop.md
@@ -145,26 +145,14 @@ drops first when it is concentrating on code, and a shipped feature filed under
 docker compose exec backend bench --site frontend run-tests --module benchpress.tests.test_deploy_manager
 ```
 
-Three rules, and each one has cost a real run.
+Two rules, and each one has cost a real run.
 
 **Never `--app benchpress`.** The whole suite is CI's job. It takes 60 seconds
-against the live site, one of its tests fails there for good (the fence list
-below), and the rest of the run does not repeat. Three runs of the same tree gave
+against the live site and does not repeat. Three runs of the same tree gave
 one failure, then four, then one: `test_api` blew a 600 ms budget at 1,473 ms
 under whole-suite load, and `test_credit_guard` got a Docker 404 that it does not
 get on its own. Scoped `--module` runs of the same tests were stable every time.
 A red whole-suite run here tells a loop nothing it can act on.
-
-**Never name a fenced module.** `test_lease` is the only one left.
-`test_a_restarted_bench_gets_a_fresh_deadline_and_is_not_reclaimed` errors here
-because `guard._enforce` holds credits against the caller instead of the bench
-owner, so an admin restarting a bench they do not own is priced against the wrong
-account. That is a production bug, not a test one, and no phase spec works around
-it. CI stays green only because a fresh `Administrator` on `test_site` collects
-the signup grant. Tracked as
-[#202](https://github.com/Venkateshvenki404224/benchpress/issues/202) — item 9's
-spec fixes it and deletes this entry. A phase that must touch `test_lease` gates
-on CI alone, and its phase spec says so.
 
 `test_credit_sweep`, `test_demo_data` and `test_signup` were fenced here until
 [#203](https://github.com/Venkateshvenki404224/benchpress/issues/203) fixed them,
--- a/benchpress/api.py
+++ b/benchpress/api.py
@@ -25,9 +25,10 @@ from benchpress.credits.guard import (
 	build_charge,
 	cap_builds_per_day,
 	cap_devices,
+	cap_size_tier,
+	deploy_lease_cost,
 	instance_lease_cost,
 	lab_owner,
-	payload_lease_cost,
 	require_balance,
 	requires_admission,
 )
@@ -187,7 +188,7 @@ def _counts_by_bench(doctype: str, column: str, bench_names: list[str]) -> dict[
 
 
 @frappe.whitelist()
-@requires_admission(cost=payload_lease_cost)
+@requires_admission(cost=deploy_lease_cost, caps=(cap_size_tier,))
 def create_bench(data: str) -> dict:
 	require_app_user()
 	from benchpress.benchpress.doctype.bench_instance import get_instance_id
@@ -197,6 +198,10 @@ def create_bench(data: str) -> dict:
 	lab_name = data.get("lab")
 	if not lab_name:
 		frappe.throw(_("Lab is required to create a bench."))
+
+	size = data.get("instance_size")
+	if size and not frappe.db.exists("Instance Size", size):
+		frappe.throw(_("There is no instance size named '{0}'.").format(size))
 
 	lab = frappe.get_cached_doc("Lab", lab_name)
 	requested_site_name = data.get("site_name") or data.get("site")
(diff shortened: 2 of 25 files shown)


Venkateshvenki404224:
## Phase 2 — the tier owns the knobs (`25be8fa`)

`Instance Size` now carries every density knob, the size binds at deploy and is never copied, and `create_bench_container` reports what it actually applied.

### What changed

- **`Instance Size`** gains `pids_limit`, `iops_limit`, `bps_limit`, `disk_limit` (Int) and `include_code_server` (Check). Nothing reads `include_code_server` yet — phase 3 is that.
- **`Bench Instance`** gains `instance_size`, written by the deploy and read by billing.
- **`Lab.apply_instance_size` is deleted.** That copy ran in `Lab.validate`, so a size edited in Desk reached no existing Lab until each one was re-saved. `Lab.memory_limit` / `cpu_cores` stay as the hand-typed fallback.
- **`docker_manager`** owns the clamp and the probe. `nano_cpus` is clamped to `os.cpu_count()`. `disk_limit` ships behind a positive probe — overlay2 **and** xfs, cached per process — because `--storage-opt size=` is accepted and silently ignored anywhere else. Probe fails while `disk_limit` is set: log and skip, never pass the flag. `create_bench_container` returns the applied and skipped limits rather than a bare container id, which is what makes that rule writable.
- **`config.size_for_instance(bench)`** is the billing-side read, and `lease.cost_of_bench` prices from it.
- **`lab_detail.get_lab`** reports the resolved size. Its `memory_limit` / `cpu_cores` were only ever right because the deleted copy kept them so.

### How to verify

Tests, reading `$?`: `test_docker_manager` (11 new), `test_credits`, `test_admission`, `test_lease`, `test_ingress`, `test_deploy_manager`, `test_lifecycle`, `test_image_cache`, `test_credit_config`, `test_lab_templates`, `test_api` — all exit 0. `uvx pre-commit@4.3.0 run --all-files` clean.

Live, on staging. A throwaway size was created, a Lab was pointed at it and saved, **then the size was edited in Desk** and a bench deployed from that Lab **without re-saving it**:

```
docker inspect <cid> --format '{{.HostConfig.PidsLimit}} {{.HostConfig.NanoCpus}} {{.HostConfig.Memory}}'
128 2000000000 671088640
StorageOpt=null ReadIOps=[{"Path":"/dev/xvda","Rate":500}] WriteBps=[{"Path":"/dev/xvda","Rate":20971520}]
```

640 MB is the size's edited value, and the Lab still stored `512m` — the staleness bug is closed. `NanoCpus` is 2, clamped from the tier's 4 on this 2-core host. The Deploy Log reads:

```
limits P2 Tier Verify — memory=640m nano_cpus=2000000000 pids_limit=128 iops=500 bps=20971520
limit skipped — disk_limit 4g — this host enforces no writable-layer quota (overlay2 on xfs required)
```

Pricing follows the bench, not the Lab: with the bench recording the 2.0x size and the Lab re-pointed at Small, `cost_of(lab, plan)` gave 5.0 and `cost_of_bench(bench, lab, plan)` gave 10.0.

All fixtures were removed — bench torn down, lab and size deleted, image tag untagged. `enable_credits` still reads 0.

Venkateshvenki404224:
## Phase 3 — `include_code_server` decides whether `ide-<id>` exists at all

Closes the live bug: `enable_code_server` gated `_start_code_server` and gated nothing else, so a lab with the IDE off still published a public `ide-<id>.benchpress.cloud` hostname in front of a port nothing listens on.

**What changed**

- `ingress.publish(instance_id, base_domain, ide=None)` emits the `ide-` router *and* service only when the bench resolves to an IDE. With `ide=None` it reads the flag itself, so no caller can supply one.
- `ingress.has_ide` / `lab_has_ide` / `_ide_for` — one rule, and it is a conjunction: the Lab asks for an IDE **and** the size it deployed at includes one. The recorded `Bench Instance.instance_size` wins over the Lab's current size, the same order billing prices at.
- `_routable_instance_ips` became `_routable_instances`, returning `name -> has_ide` from one join over Bench Instance × Lab. The `*/5` pass resolves the whole fleet in that read; measured **3 queries per pass with 8 benches**, unchanged by bench count.
- The deploy gate reads the same resolver, so the router and the process cannot disagree, and a bench that resolves to no IDE has its `code_server_url` cleared.
- `lab_detail.get_lab` reports the resolved flag, so the Lab screen stops offering a code-server chip and connection rows for a size that has none.

**One deviation from the spec, deliberate.** The spec described the Lab flag as an *override* of the tier. That needs a flag that can be unset; `Lab.enable_code_server` is `tinyint NOT NULL DEFAULT 1` wired to a checkbox in `NewLab.vue` and read by `LabHeader` and `ConnectionDetails`, so the override direction would mean converting a live user-facing boolean to a tri-state and migrating three SPA components. A size with the flag off is also a density statement — Nano is 512m and half a core — so overriding it upward would run an IDE on a bench that cannot host one.

**How to verify**

```bash
docker compose exec backend bench --site frontend run-tests --module benchpress.tests.test_lifecycle; echo "exit=$?"
```

Live, on a throwaway `Instance Size` with `include_code_server = 0` and a Lab whose own flag was left **on** — so the tier is what turned the IDE off:

| Check | Result |
|---|---|
| Route file | `site-<id>` only; `grep -c 'ide-'` → **0** |
| `curl https://ide-<id>.benchpress.cloud` | **404** (and 404 straight at Traefik via `--resolve`, so no router matches — not a 502 in front of a dead port) |
| Deploy log, step 10 | `Code server is disabled for this lab or its instance size — skipped` |
| `Bench Instance.code_server_url` | `NULL` |
| Two `ingress.reconcile` passes | every route file's mtime **unchanged** — the deploy and the cron write the same bytes |
| Query cost | `Questions` delta **3** per pass with 8 benches |
| Collateral, a bench that has an IDE | `ide-` route present, `302` → `200` with `<title>code-server login`; all 7 pre-existing route files byte-identical to what the new code writes |
| Lab screen | no code-server chip and no code-server rows for the no-IDE lab; both still present on `crm` |

Fixtures removed afterwards: bench torn down, lab and size deleted, image tag untagged. `BenchPress Settings.enable_credits` reads 0.

Venkateshvenki404224:
## Phase 4 — a caller may name only a size they can pay for

`5b17843` closes the last of item 9. All five checks green on this commit.

### What it adds

**The hold is priced from the size the deploy will actually use.** It was priced
from the Lab's, so a caller naming a larger size was held for the smaller one —
and the hold is the only thing in the app that can refuse a start.

- `guard.deploy_lease_cost` prices `create_bench`, `enqueue_deploy` and
  `enqueue_redeploy` at `deploy_size(**call)`: the size the call names, else the
  Lab's own chain.
- `guard.instance_lease_cost` now prices a *start* at
  `config.size_for_instance(bench)` — the size the container was given, which is
  what `metering.charge_lease` goes on to debit.
- `api.create_bench` accepts `instance_size`, refuses a name that does not exist,
  and forwards it to `lifecycle.deploy_bench(bench_name, size=...)`. The gate and
  the deploy read one chain, so what is priced is what Docker is given.

**`Credit Settings.max_size_free`** is the largest tier an account that has never
purchased may deploy at, compared by `price_multiplier`. Empty means no ceiling,
following `max_concurrent_uncredited`. It exists for one measurement: the
40-credit signup grant buys exactly two 20-credit leases, and `placement.py`
counts address slots and reads no memory, so nothing else stood between a signup
and 8 GB on a 7.7 GiB host.

**The two refusals read differently**, because an operator cannot act on one
sentence that means both:

```
ceiling        The P4 Big size is not available on a free account. Deploy at P4 Free
               or smaller, or buy credits at /frontend/credits to unlock every size.
affordability  Not enough credits: this needs 20.0 and 3.0 are available — 17.0 short.
               Top up at /frontend/credits.
```

### How to verify it

```bash
bench --site <site> run-tests --module benchpress.tests.test_credit_guard   # 32, +12 new
bench --site <site> run-tests --module benchpress.tests.test_credits
bench --site <site> run-tests --module benchpress.tests.test_admission
bench --site <site> run-tests --module benchpress.tests.test_lease
bench --site <site> run-tests --module benchpress.tests.test_api
```

Walked live on staging with credits forced on **in memory only** — the written
switch would turn credits on for every worker:

| row | outcome |
|---|---|
| credits off, ceiling set | Small, Medium, Large, and two throwaway tiers all admitted; no gate at all |
| never purchased, over the ceiling | refused on the ceiling, funded 50× past the price |
| never purchased, at the ceiling | admitted |
| purchased, affordable | admitted — a Purchase row lifts the ceiling |
| purchased, short | refused on affordability, naming the 17.0 gap |
| ceiling empty | admitted at every tier |
| fresh account, grant only, ceiling set | refused on the ceiling |
| fresh account, grant only, ceiling clear | two 20-credit leases admitted, reserved 40.0, slots 2 — the hole the ceiling closes |

One real container proved the override reaches Docker rather than only the gate.
A Lab pointing at a 768m/1-core tier, deployed naming a 1200m/2-core one:

```
Memory=1258291200  NanoCpus=2000000000  PidsLimit=111
limits P4 Big — memory=1200m nano_cpus=2000000000 pids_limit=111 iops=1000 bps=41943040
Bench Instance.instance_size = P4 Big
```

The seven benches already running were untouched, and a start is deliberately
not gated by the ceiling: this phase gates deploys.

### One thing the operator has to do

All three seeded sizes carry `price_multiplier = 1.0`, so setting `max_size_free`
today gates nothing. The ceiling follows the credit ladder on purpose — a tier's
price is a business decision, not a migration — so seeding the ladder and setting
`max_size_free` go together. That pairing is the spec's own leftover step.

PR #256, opened 2026-08-27
Implemented phase four

Venkateshvenki404224:
Phase 1 of the docker-events-listener spec: the Docker daemon's event stream reaches the database.

## What this adds

- **`benchpress/docker_events.py`** — a long-lived consumer, shaped like `credits/warden.py`. A reader thread does no Frappe work (`frappe.local` is thread-local) and appends to a queue; the main loop parks each incident with a due timestamp and judges it after the settle window.
- **`Bench Event`** — a log-shaped doctype, tenant-scoped with both permission halves, 30-day retention in `default_log_clearing_doctypes`.
- **`bench_event_settle_seconds`** on BenchPress Settings, default 15. It must exceed `stop_grace_seconds`.
- **`docker-events`** compose service in the base file and the prod overlay (the base file's `deploy.restart_policy` is a Swarm field `up` ignores).

## Why it decides nothing at event time

`stop_bench` stops the container and commits `Stopped` afterwards, so a `die` arrives while the row still says `Running`. A listener reading the database at event time alerts on every ordinary stop.

Two corrections the live system forced:

1. **`Stopped` alone does not mean "the platform asked".** `stats_collector._stop_if_dead` also writes it, for exactly the benches that died — and it beat the 15-second window on the first drill. It stamps the health it found on the way, so `Stopped` + `Unhealthy`/`Unknown` is a death and `Stopped` + the health it had while serving is a stop.
2. **A start now clears `container_health`** (`lifecycle.running`). A stale `Unhealthy` surviving a restart would otherwise read as a death for a poll interval.

## How to verify it

    cd /home/ubuntu/benchpress_devops
    docker compose up -d --no-deps docker-events
    </dev/null docker compose exec -T backend bench --site frontend execute benchpress.docker_events.heartbeat_value

Measured on staging with drill bench `31bc285092e6d65263d849f9549defea`:

| Step | Result |
|---|---|
| `docker kill` the bench | one row, `bench_died`, `error`, exit 137, action `die`, `occurred_at` = the daemon's time; one `Alert` to the owner |
| scratch container OOM under the bench's labels | one row, `oom_killed`, exit 137 — not an `oom_killed` **and** a `bench_died` |
| stop the bench through `api.bench_action` | `events_seen` rises, row count unchanged |
| non-owner `BenchPress User` | lists 0 rows, `frappe.client.get` raises; the owner lists and reads its own |

`run-tests --app benchpress` 857 + 218 tests, exit 0. `uvx pre-commit@4.3.0 run --all-files` clean.

Traefik was not restarted, no certificate was issued, and the host is untouched.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -127,7 +127,10 @@ jobs:
       - name: Setup
         run: |
           pip install frappe-bench
-          bench init --skip-redis-config-generation --skip-assets --python "$(which python)" ~/frappe-bench
+          # --frappe-branch, for the same reason vpn_management is pinned below: bench init
+          # defaults to frappe's develop branch, so the suite tracked a branch this app does
+          # not ship on and broke on upstream changes no pull request here contained.
+          bench init --frappe-branch version-16 --skip-redis-config-generation --skip-assets --python "$(which python)" ~/frappe-bench
 
       # benchpress declares required_apps = ["vpn_management"], a public repo,
       # so install-app benchpress only resolves if the app is already on the bench.
--- a/benchpress/benchpress/doctype/bench_event/bench_event.json
+++ b/benchpress/benchpress/doctype/bench_event/bench_event.json
@@ -1,0 +1,124 @@
+{
+ "actions": [],
+ "allow_rename": 0,
+ "autoname": "hash",
+ "creation": "2026-08-27 00:00:00.000000",
+ "doctype": "DocType",
+ "editable_grid": 0,
+ "engine": "InnoDB",
+ "field_order": [
+  "bench",
+  "event_type",
+  "severity",
+  "column_break_event",
+  "occurred_at",
+  "docker_action",
+  "exit_code",
+  "detail"
+ ],
+ "fields": [
+  {
+   "fieldname": "bench",
+   "fieldtype": "Link",
+   "in_list_view": 1,
+   "in_standard_filter": 1,
+   "label": "Bench",
+   "options": "Bench Instance",
+   "reqd": 1,
+   "search_index": 1
+  },
+  {
+   "fieldname": "event_type",
+   "fieldtype": "Select",
+   "in_list_view": 1,
+   "in_standard_filter": 1,
+   "label": "Event Type",
+   "options": "bench_died\noom_killed",
+   "reqd": 1
+  },
+  {
+   "default": "error",
+   "description": "The SPA notification badge reads this, so it is a contract and not a label.",
+   "fieldname": "severity",
+   "fieldtype": "Select",
+   "in_list_view": 1,
+   "in_standard_filter": 1,
+   "label": "Severity",
+   "options": "error\nwarning\ninfo"
+  },
+  {
+   "fieldname": "column_break_event",
+   "fieldtype": "Column Break"
+  },
+  {
+   "description": "When the Docker daemon reported it, not when the row was written.",
+   "fieldname": "occurred_at",
+   "fieldtype": "Datetime",
+   "in_list_view": 1,
+   "label": "Occurred At",
+   "reqd": 1,
+   "search_index": 1
+  },
+  {
+   "description": "The raw event action, e.g. die or health_status: unhealthy.",
+   "fieldname": "docker_action",
+   "fieldtype": "Data",
+   "label": "Docker Action",
+   "read_only": 1
+  },
+  {
+   "fieldname": "exit_code",
+   "fieldtype": "Int",
+   "label": "Exit Code",
+   "read_only": 1
+  },
+  {
+   "fieldname": "detail",
+   "fieldtype": "Small Text",
+   "label": "Detail",
+   "read_only": 1
+  }
+ ],
+ "grid_page_length": 50,
+ "index_web_pages_for_search": 1,
+ "links": [],
+ "modified": "2026-08-27 00:00:00.000000",
+ "modified_by": "Administrator",
+ "module": "Benchpress",
+ "name": "Bench Event",
+ "owner": "Administrator",
+ "permissions": [
+  {
+   "delete": 1,
+   "email": 1,
+   "export": 1,
+   "print": 1,
+   "read": 1,
+   "report": 1,
+   "role": "System Manager",
+   "share": 1
+  },
+  {
+   "delete": 1,
+   "email": 1,
+   "export": 1,
+   "print": 1,
+   "read": 1,
+   "report": 1,
+   "role": "BenchPress Admin",
+   "share": 1
+  },
+  {
+   "export": 1,
+   "read": 1,
+   "report": 1,
+   "role": "BenchPress User"
+  }
+ ],
+ "row_format": "Dynamic",
+ "rows_threshold_for_grid_search": 20,
+ "sort_field": "occurred_at",
+ "sort_order": "DESC",
+ "states": [],
+ "track_changes": 0
+}
--- a/benchpress/benchpress/doctype/bench_event/bench_event.json
+++ b/benchpress/benchpress/doctype/bench_event/bench_event.json
@@ -33,7 +33,7 @@
    "in_list_view": 1,
    "in_standard_filter": 1,
    "label": "Event Type",
-   "options": "bench_died\noom_killed",
+   "options": "bench_died\noom_killed\nbench_unhealthy\nbench_healthy",
    "reqd": 1
   },
   {
--- a/benchpress/benchpress/doctype/bench_event/bench_event.py
+++ b/benchpress/benchpress/doctype/bench_event/bench_event.py
@@ -1,0 +1,8 @@
+# Copyright (c) 2026, Venkatesh and contributors
+# For license information, please see license.txt
+
+from frappe.model.document import Document
+
+
+class BenchEvent(Document):
+	pass
--- a/benchpress/benchpress/doctype/benchpress_settings/benchpress_settings.json
+++ b/benchpress/benchpress/doctype/benchpress_settings/benchpress_settings.json
@@ -18,6 +18,7 @@
   "bench_subnet_base",
   "bench_bridge_count",
   "bench_slots_per_bridge",
+  "bench_event_settle_seconds",
   "resource_section",
   "container_memory_limit",
   "column_break_resource",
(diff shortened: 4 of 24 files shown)


Venkateshvenki404224:
Two CI fixes rode along with phase 1, each in its own commit.

**`fix(ci)`: frappe is pinned to `version-16` in the test bench.** `bench init` with no branch takes frappe's `develop`. The `Server` job was green at 16:40 today and red at 17:40 on two `test_workspace` fixture assertions, and the only thing between the two runs was frappe/frappe#41636 ("first-class modules") landing on develop at 17:38. The suite was testing a branch this app does not ship on — `entry.py` builds `version-16` and the deployment runs it — so the pin is the same reasoning the pinned `vpn_management` commit beside it already carries. Both workspace tests pass on `version-16`.

**`fix(events)`: the listener's `frappe.db.commit()` carries `# nosemgrep`.** `frappe-manual-commit` blocks the runner and the local pre-commit hooks do not include semgrep, so it passed lint on the dev host. The commit is required: a service has no request boundary to end its transaction, and `_start_fresh` rolls back at the top of the next tick.

Also worth knowing for the next phase: `pre-commit run --all-files` runs on the files git knows about, so a new **untracked** file is skipped entirely and passes a lint the runner then fails. Stage before believing a green run.

All five checks green on `3356e4a`. Re-verified live on that commit: `_stamp(1787851412)` → `2026-08-27 22:53:32` (the daemon's epoch in the site's timezone), and a scratch `die` labelled with a bench name that has no row moved `events_seen` 0 → 1 and `orphans` 0 → 1 while writing no row. Traefik's `StartedAt` is byte-identical to the loop's preflight snapshot, all seven routed benches answer 200, and the control plane answers pong.

Venkateshvenki404224:
## Phase 2 — a bench tells the truth about itself

`get_container_health` returned `Healthy` for every container whose state was `running`. A bench runs `tail -f /dev/null` as PID 1 with `bench serve` in a detached `screen` session, so a dead site read `Healthy` for ever while Traefik kept routing to it.

**What this adds**

- A Docker healthcheck on `/api/method/ping` for every new bench container, built by `docker_manager.bench_healthcheck()` from six new `BenchPress Settings` fields: `enable_bench_healthcheck`, `bench_health_interval_seconds` (30), `bench_health_timeout_seconds` (5), `bench_health_retries` (3), `bench_health_start_period_seconds` (600), `stats_poll_max_benches` (50).
- `health_status` joins `EVENT_FILTERS`, and `bench_unhealthy` / `bench_healthy` join `Bench Event`. A health incident also freshens `container_health` and `last_health_check`, so the field is right in seconds instead of after a poll interval.
- `get_container_health` returns Docker's verdict, and falls back to the old run-state rule for a container that has no healthcheck, so a bench created before this phase keeps answering what it always did.
- `stats_collector` bounds its poll and takes the stalest bench first.
- `patches/backfill_bench_health` stamps `container_health` from the live container and writes the new settings defaults. Four `Stopped` benches that claimed `Healthy` now read `Unhealthy`.

**Two things the phase spec did not ask for, both caused by this change**

- `_stop_if_dead` now acts on the container's run state, not the health label. Without this, `Unhealthy` — which now also means "the site went quiet inside a running container" — would make the poll **stop the bench**, so a 90-second gunicorn hiccup would take the tenant's work with it.
- A recovery with no recorded failure behind it writes no row, because the first verdict of every deploy is `starting -> healthy`.

**Verified live on staging**

| Step | Result |
|---|---|
| The gap, before | Site killed on a drill bench: `docker inspect` → `running null`, `curl localhost:8000` → connection refused, `get_container_health` → **Healthy**, row → `Running / Healthy` |
| Healthcheck on the container | `Interval` **30000000000**, not 30. `StartPeriod` 600000000000 |
| The same kill, after | `unhealthy` with `FailingStreak` 3 at 80 s. One `bench_unhealthy` row, `docker_action` `health_status: unhealthy`. Row → `Unhealthy`, `last_health_check` stamped 15 s later by the settle window. Owner alerted |
| Recovery | `serve.sh` → `healthy` in 6 s → one `bench_healthy` row, `severity` `info`, row back to `Healthy` |
| A deploy does not trip it | Golden restore: container start → serving 15 s. Golden restore off, site created in the container: **45 s**, health held at `starting` and reached `healthy` at 48 s, never `unhealthy`. The seven-app worst case is 202 s against a 600 s start period |
| The switch | `enable_bench_healthcheck = 0`, redeploy → `Config.Healthcheck` is absent, and `get_container_health` answers `Healthy` from run state |
| Tests and lint | 859 + 251 tests pass. All 12 pre-commit hooks pass. The frappe semgrep rules and `r/python.lang.correctness` are clean |

Traefik was not restarted (`StartedAt` unchanged), no certificate was issued, and all seven routed benches answer 200.

**Known gap, narrowed and not closed.** A bench that reads `Unhealthy` while `Running` and is then stopped on request still settles as `bench_died`, because `_unasked_for` reads `Stopped` plus `Unhealthy` as "the poll found it dead". The window needs a site that was already broken when the owner stopped the bench, and the `bench_unhealthy` row sits in front of it in the same feed.

**Still out of scope.** Traefik routes to an unhealthy bench — Docker's verdict is invisible to it. Phase 3 gives Traefik its own probe.

Venkateshvenki404224:
## Phase 3 — routing follows health

Every bench route file now carries Traefik's **own** `loadBalancer.healthCheck`, one per service, each probing the path that service actually serves. Docker's healthcheck (phase 2) feeds the event stream and ejects nothing: Traefik reads the file provider and has no socket to hear the daemon's verdict on. These are two mechanisms with two owners.

- `ingress.publish` — `healthCheck` on both the `site-` and `ide-` services. Site probes `/api/method/ping`, IDE probes `/healthz`.
- `BenchPress Settings` — `traefik_health_interval_seconds` (10) and `traefik_health_timeout_seconds` (3), plus a patch that writes the defaults over unset **and** over a stored zero.
- `TestServiceHealthCheck` — 7 tests, asserting each path separately, because a copy-paste that gives both the same path passes any test that only checks "a healthCheck exists".

The paths differ on purpose. A shared `/api/method/ping` would eject a working code-server every time gunicorn hiccuped; a shared `/healthz` would eject nothing at all, because code-server answers it while Frappe is dead.

The probe rate is a setting rather than a constant because it is a fleet-wide load, not a preference: Traefik probes every server on every service, so 100 benches at a 10-second interval is 20 full Frappe requests a second against a two-worker gunicorn.

### How to verify

```bash
# 1. Two health checks per route file, none on the protected two.
for f in config/traefik/generated/dynamic/*.yml; do echo "$f $(grep -c healthCheck: $f)"; done

# 2. Every healthy bench is unaffected.
curl -s -o /dev/null -w '%{http_code}\n' --resolve $B.benchpress.cloud:443:127.0.0.1 \
  https://$B.benchpress.cloud/api/method/ping     # 200

# 3. Kill the site, wait ~11 s, and read the discriminator.
docker exec $C sh -c "sudo -u $USER screen -X -S benchweb quit"
curl ... /api/method/ping                          # 503, not 502
docker logs --since 60s benchpress_traefik | grep 'Health check failed'

# 4. Recover; no file is rewritten and Traefik is not restarted.
docker exec $C bash /opt/benchpress/scripts/serve.sh
```

### Measured live

| | |
|---|---|
| Ejection | site killed 19:00:51 → **503 at 19:01:01** (11 s) |
| The discriminator, both halves in one access log | `DownstreamStatus:502 / OriginStatus:502` before, `DownstreamStatus:503 / OriginStatus:0` after — dialled-and-failed vs never-dialled |
| The IDE, with the site ejected | **200** throughout; no health-check failure ever named the `ide-` service |
| Recovery | `serve.sh` 19:01:17 → **200 at 19:01:30** (13 s), route file byte-identical (same mtime, same md5), no proxy restart |
| Healthy benches after the reconcile | all seven **200 on site and IDE**, 2 health checks each |
| Negative control | five non-Running benches still own **no** route file — ejection is not withdrawal |

Traefik's `StartedAt` is unchanged at `2026-08-25T18:13:06.342173753Z`. No certificate issued, no `certResolver` added to any bench router, `wildcard-anchor.yml` untouched.

CI green on `35aae4f` — 5/5.

Venkateshvenki404224:
Phase 4 — convergence and the feed. The listener can no longer die quietly, a restart re-reads the world instead of trusting the daemon buffer, and bench events reach the tenant they are about.

**What this adds**

- `docker_events.reconcile()` — re-reads health for every `Running` bench and runs at the top of every connection. The daemon keeps 255 events (74 s on this host), so `since=<disconnect>` is a head start, never a guarantee. Returns `{checked, changed, recorded}`.
- A `*/5` cron enqueues it on `queue-long`, so a listener that stays down degrades to the convergence BenchPress had before this spec rather than to silence.
- The event is gated on the **last event recorded for the bench**, not on a change to `container_health`. `stats_collector` writes that field too and records nothing, so a poll winning the race would otherwise silence the pass entirely.
- `diagnostics` gains an eleventh row, `docker_events`. A stale or missing heartbeat fails and names the service to start; the threshold is derived from the tick rather than being a second constant.
- `overview._activity` merges `Bench Event`, already scoped by `bench_event_query_conditions`. Each event type gets its own sentence — reusing `_deploy_event` would render a dead bench as "is deploying".

**How it was verified, live on staging**

1. *The row goes red only when the listener stops.* Eleven rows each time: `PASS` (0 s ago) → stop → `FAIL` after 98 s naming `docker compose up -d docker-events` → start → `PASS` (1 s ago). Checked in that order, because a row that never goes red proves nothing.
2. *Reconciliation, not replay.* Listener stopped, a drill bench's site killed, Docker's `unhealthy` verdict at 19:42:39. By 19:44:34 the buffer held 256 events with the oldest at 19:43:19 — the verdict was **40 s older than anything still replayable**, and the only events left for that container were healthcheck `exec_*` noise. On start the pass wrote `Unhealthy` and a `bench_unhealthy` row stamped at the pass.
3. *An intentional stop produces nothing.* Same window, bench stopped through `lifecycle.stopped`: **zero** `Bench Event` rows on restart.
4. *The cron is a real net.* Listener left `exited` for the whole test; the `*/5` pass converged the field and the event at t+237 s without it.
5. *The owner sees it.* Signed in as the bench owner, the panel reads "bench-crm is not responding" / "bench-crm is responding again" with the badge counting 3 — not "is deploying".
6. *A second tenant sees none of it.* `dk-phase1-tenant` owns one bench whose events are `bench_died`/`oom_killed`; `dk-phase2-tenant` owns one whose events are `bench_unhealthy`/`bench_healthy`. Each panel carried exactly its own set and nothing of the other's.

`bench run-tests --app benchpress` 860 + 275, exit 0. `vitest run` 184. `pre-commit` 12 hooks.

Traefik was never restarted — `StartedAt` is still `2026-08-25T18:13:06.342173753Z`. All seven routed benches serve 200.

**One thing to know:** `docker compose up -d <any service>` re-runs the `configurator` one-shot, which rewrites `common_site_config.json` from its heredoc. That heredoc had no `workers` key, so the rewrite dropped the `stops` queue declaration and left `queue-stops` in a restart loop. Fixed in the parent repo ([benchpress_devops 99df896](https://github.com/Venkateshvenki404224/benchpress_devops/commit/99df896)) by putting the key in the heredoc, and the live config restored.

PR #257, opened 2026-08-28
Implemented phase two

Venkateshvenki404224:
Phase 1 of `two-tier-rate-limiting`: three routers per bench, each with its own
rate limit keyed on the real client.

## What changed

- `benchpress/ingress.py` — `publish` emits `assets-<id>`, `site-<id>` and
  `ide-<id>`. Every router carries one `rateLimit` middleware of its own name,
  defined in the same file that references it. The assets router states a
  higher `priority` than the site router.
- `Instance Size` — `rate_average` (20), `rate_burst` (60) and `inflight_limit`
  under a new **Request Limits** section. `inflight_limit` is inert until
  phase 2 and lands now so there is one schema change, not two.
- `benchpress/patches/seed_route_rate_limits.py` — backfills rows that predate
  the fields, because Traefik reads `average: 0` as no limit at all.

No migration is needed for the route files: the `*/5` reconcile rewrites every
one from the database.

## How to verify

```bash
docker compose exec queue-long bench --site frontend execute benchpress.ingress.reconcile
cat config/traefik/generated/dynamic/<bench-id>.yml
```

Three routers, three middlewares, `depth: 1` on each, no `inFlightReq`.

Live, on staging — all seven benches:

| check | result |
|---|---|
| page and a real asset | both 200 |
| IDE hostname | 200 |
| control plane | 200 |
| `grep middlewares config/traefik/generated/traefik.yml` | no match |

**The limit engages.** The spec's step 6 (80 serial requests for missing
assets) returned 80 × 404 and no 429 — a serial loop against a 226 ms render is
4 rps, which never drains a 20/s bucket, so the arithmetic is against that
check rather than the implementation. Proven instead by lowering `Small` to
`average: 1, burst: 2`, reconciling, and firing 12 serial requests:

```
404 404 429 404 429 429 429 429 429 404 429 429
```

Two on the burst, then 429s with 404s as the bucket refills at 1/s. The tier
was restored to 20/60 in the same session and every route file re-checked.

**`depth: 1` is not spoofable.** Under the same lowered tier, requests carrying
`X-Forwarded-For: 1.2.3.4` shared the bucket of requests without one, instead
of getting a fresh one.

**The split holds.** With the assets bucket fully spent, `GET /` still returned
200 — the render router has its own bucket, which is what makes phase 2's
in-flight cap safe to add.


Venkateshvenki404224:
**Phase 2 — the render router gets a per-tier in-flight cap**

`site-<id>` now carries a second middleware, `<id>-inflight`, at
`min(Instance Size.inflight_limit, INFLIGHT_CEILING)` with `INFLIGHT_CEILING = 24`.
`assets-<id>` and `ide-<id>` carry none, and no Traefik entrypoint was touched.

A tier with no `inflight_limit` publishes no middleware rather than a default, so
this ships inert: every live tier reads 0 today and the five route files are
unchanged until an operator seeds the ladder.

**How to verify**

```
```

`TestInFlightCap` covers the chain (rate limit first, cap second, render router
only), the tier amount, the clamp, the unseeded case and the absence of
`sourceCriterion`.

Verified live on staging, then restored:

- Tier set to 999 → every route file wrote `amount: 24`, on `site-` only.
- Tier set to 4 → six concurrent renders returned `429 429 200 200 200 200`, and a
  single request straight afterwards returned 200 in 0.24 s. Traefik logs the two
  refusals on `site-...@file` with `OriginStatus: 0` — refused at the router,
  never reaching the bench.
- A browser tab on `/login` under the cap of 4: all 11 same-host subresources
  200/204, zero 429. Only `/login` and `/website_script.js` land on the render
  router, exactly as phase 1's split intends.
- A code-server session stayed live through ~4.5 minutes of repeated saturation
  and opened a file over its WebSocket while two renders were being 429'd.
  `staging.benchpress.cloud` answered 200 throughout.
- Health checks: cap of 1 held saturated for 35 s (656 of 809 requests refused)
  across three 10 s probe intervals produced zero 503s and no server ejection —
  probes are issued from the service and never enter the router chain.
- Traefik was never restarted (`restarts=0`, up since Aug 25); `dynamic.yml` and
  `wildcard-anchor.yml` are untouched.

All three tiers were restored to `inflight_limit: 0` and the fleet reconciled back
to no `inFlightReq` anywhere. Seeding the 4/8/12/16 ladder stays the operator's
step, per the spec README.

PR #258, opened 2026-08-28
Implemented phase three

Venkateshvenki404224:
Phase 1 of the one-reconcile-loop spec: the upstream fix that decides how much work phase 3's orphan reporting will find.

## What changed

`lifecycle.torn_down` swallowed four failures — the stop, the container removal, the site-database drop and the route-file delete — and then wrote `Draft` unconditionally. A teardown whose container removal failed reported exactly what one that worked reported: nothing.

- Each removal now returns `gone` or `failed: <reason>`, and `torn_down` returns the five results.
- A teardown that left something behind writes it to the Error Log, so no caller can forget to.
- The swallowing stays. An instance must not be left describing resources it no longer has.
- Teardown now calls `remove_bench_peer`. Only `api.destroy_bench` and the failed-deploy cleanup freed the VPN allocation before, so every *reaped* bench held its tunnel address forever. `wg_ip` is cleared with it — the pool has taken the address back.
- Three docstrings promised a per-bench volume. There is none; `create_bench_container` mounts none on purpose.

`FakeContainer` gained `stop_refusal` / `remove_refusal`, mirroring `start_refusal`, so a refused removal is scripted through the fake rather than mocked.

## How to verify

Scoped tests:

```
docker compose exec backend bench --site frontend run-tests --module benchpress.tests.test_lifecycle
docker compose exec backend bench --site frontend run-tests --module benchpress.tests.test_vpn_adapter
```

Against the live stack, a throwaway bench on `client-frappe-16`:

| step | result |
|---|---|
| baseline allocated `IP Allocation` rows | 16 |
| after deploy (PEER-00115, 172.27.0.18) | 17 |
| after `reap_bench` | **16** — the leak is closed |
| container, route file, site database `_e4014d6da5aa788d` | all gone; `dynamic.yml` and `wildcard-anchor.yml` untouched |
| clean teardown | no Error Log row |
| redeploy of the torn-down bench | Running, fresh peer PEER-00116, address 172.27.0.18 reissued from the pool |
| teardown after `docker rm -f` on its container | `container: failed: 404 ... No such container: 5a0ae822…` in the Error Log, bench still reached `Draft`, allocation still returned to 16 |

The throwaway bench and its Error Log rows were deleted afterwards; the fleet is back to its baseline.

### DIFF
--- a/benchpress/api.py
+++ b/benchpress/api.py
@@ -342,9 +342,8 @@ def _stop_bench(bench) -> dict:
 def _delete_bench(bench) -> dict:
 	"""Remove an instance and everything it owns, then the row itself.
 
-	Container, volume, site database and metering session go through `lifecycle.torn_down`,
-	the one teardown path, where every removal is best-effort. Only what it does not cover is
-	left here.
+	`lifecycle.torn_down` is the one teardown path; the other sites' databases, and a peer it
+	reported it could not remove, are what is left here.
 	"""
 	from benchpress.vpn_adapter import remove_bench_peer
 
--- a/benchpress/benchpress/doctype/benchpress_settings/benchpress_settings.json
+++ b/benchpress/benchpress/doctype/benchpress_settings/benchpress_settings.json
@@ -36,7 +36,8 @@
   "credits_section",
   "enable_credits",
   "reconcile_section",
-  "orphan_grace_minutes"
+  "orphan_grace_minutes",
+  "deploy_log_cap"
  ],
  "fields": [
   {
@@ -238,6 +239,13 @@
    "fieldname": "orphan_grace_minutes",
    "fieldtype": "Int",
    "label": "Orphan Grace Minutes"
+  },
+  {
+   "default": "50",
+   "description": "Deploy Log rows kept per bench. Frappe's own sweep clears them at seven days; this is the second bound, for a bench redeployed in a loop inside one.",
+   "fieldname": "deploy_log_cap",
+   "fieldtype": "Int",
+   "label": "Deploy Log Cap"
   }
  ],
  "grid_page_length": 50,
--- a/benchpress/benchpress/doctype/benchpress_settings/benchpress_settings.json
+++ b/benchpress/benchpress/doctype/benchpress_settings/benchpress_settings.json
@@ -34,7 +34,9 @@
   "code_server_version",
   "stop_grace_seconds",
   "credits_section",
-  "enable_credits"
+  "enable_credits",
+  "reconcile_section",
+  "orphan_grace_minutes"
  ],
  "fields": [
   {
@@ -224,13 +226,25 @@
    "fieldname": "enable_credits",
    "fieldtype": "Check",
    "label": "Enable Credits"
+  },
+  {
+   "fieldname": "reconcile_section",
+   "fieldtype": "Section Break",
+   "label": "Reconciliation"
+  },
+  {
+   "default": "15",
+   "description": "A container younger than this is never named an orphan: a deploy creates the container before it writes the row.",
+   "fieldname": "orphan_grace_minutes",
+   "fieldtype": "Int",
+   "label": "Orphan Grace Minutes"
   }
  ],
  "grid_page_length": 50,
  "index_web_pages_for_search": 1,
  "issingle": 1,
  "links": [],
- "modified": "2026-08-27 17:20:00.000000",
+ "modified": "2026-08-28 06:00:00.000000",
  "modified_by": "Administrator",
  "module": "Benchpress",
  "name": "BenchPress Settings",
--- a/benchpress/benchpress/doctype/benchpress_settings/benchpress_settings.py
+++ b/benchpress/benchpress/doctype/benchpress_settings/benchpress_settings.py
@@ -21,7 +21,7 @@ class BenchPressSettings(Document):
 			return
 
 		frappe.enqueue(
-			"benchpress.ingress.reconcile",
+			"benchpress.reconcile.run",
 			queue="long",
 			job_id="reconcile_instance_routes",
 			deduplicate=True,
--- a/benchpress/credits/reaper.py
+++ b/benchpress/credits/reaper.py
@@ -4,9 +4,10 @@
 """The reaper: stopped is free, but not forever.
 
 "Stopped costs nothing" is only honest if a stopped instance eventually stops costing *us*
-something too — a container, a volume and a site database sit on disk whether or not anything is
-running in them. So an instance that has sat `Stopped` past `reap_after_days` is torn down, with
-an email two days before so nobody loses work to a rule they had forgotten.
+something too — a container and a site database sit on disk whether or not anything is running in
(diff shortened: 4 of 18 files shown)


Venkateshvenki404224:
## Phase two — the direction nothing could see

`list_benches()` and `compare()` exist. **This phase removes nothing.** Phase 3 acts on what it names.

### What it adds

- `docker_manager.list_benches()` — the whole `benchpress.managed=true` fleet in one `containers.list()`, carrying `{id, name, bench_name, status, health, created}`. `all=True`, because an orphan that exited still holds its writable layer. `_health_verdict` was lifted out of `get_container_health` so a fleet read gets health without a second inspect.
- `benchpress/reconcile.py` — `compare(rows, containers) -> {orphan_containers, in_grace, missing_containers}`. It reads neither side; both arrive as arguments, so the events consumer and the `*/5` loop cannot end up with two definitions of "orphan" that undo each other.
- The grace window lives **inside** `compare`. `_deploy_bench` creates the container before it writes `container_id`, so every deploy passes through a state that looks exactly like an orphan. New setting `BenchPress Settings.orphan_grace_minutes`, seeded to 15 to match `admission_repair.CLAIM_GRACE_MINUTES`.
- `mariadb_manager.list_site_databases()` — the database port's last open edge, closed as a read.
- The managed label now has one definition instead of a literal in `create_bench_container` and a copy in `docker_events`.

### How to verify it

```bash
docker compose exec backend bench --site frontend run-tests --module benchpress.tests.test_reconcile; echo "exit=$?"
docker compose exec backend bench --site frontend run-tests --module benchpress.tests.test_docker_manager; echo "exit=$?"
```

Then look at the live drift without acting on it:

```python
from benchpress import docker_manager, reconcile
import frappe
rows = frappe.get_all("Bench Instance", fields=["name", "container_id", "status"])
print(reconcile.compare(rows, docker_manager.list_benches()))
```

### Measured against the live fleet

| | |
|---|---|
| `list_benches()` | **45.5 ms** for 11 containers (1.02 ms per inspect x 11 = 11.2 ms, but one call, not twelve) |
| drift today | **0 orphan containers, 0 in grace, 0 missing** — the one row with no container is `Draft` with a null `container_id`, correctly not drift |
| schemas | 15, of which **4 have no `Bench Site` row**: two leaked site databases, one privilege-drill artefact, and `backups`. Reported, never dropped. |

### The grace window, against a real deploy in flight

The test that matters: get this wrong and phase 3 destroys every deploy in flight. A real redeploy of `client-frappe-16`, with `compare` polled every 0.25 s throughout:

```
polls 595, elapsed 186.9 s, final_status Running
missing_seen  {bench: {"status": "Deploying", "first_seen_s": 18.1}}
in_grace_seen {ec0e8cf4...: {"status": "running", "first_seen_s": 23.8}}
orphans_seen  {}
new_container_ever_called_orphan  False
```

The window is real and **5.7 seconds wide** here — at t=18.1 s the row still named the container the redeploy had just removed, and at t=23.8 s the new container existed while no row named it. `orphan_containers` was empty for all 595 polls. The throwaway bench was then deleted and every count returned to its pre-test value.

`test_the_fleet_costs_one_call` patches `containers.get` to raise, so a future per-container implementation fails a test rather than a stopwatch.

Venkateshvenki404224:
## Phase 3 — the loop, and it re-checks its own work

`benchpress/reconcile.py` now owns the `*/5` pass. `ingress.reconcile` and
`ingress.enqueue_route_reconcile` are gone; `_routable_instances` is public as
`routable`, and `directory_mounted()` joins `published()` / `protected_present()` so the
loop never learns `TRAEFIK_DYNAMIC_DIR` or `PROTECTED_ROUTE_FILES`.

`run()` reports a count for each of six steps:

```
{"bridges": {"attached": {}, "missing": {}},
 "routes": {"anchored": false, "written": 7, "deleted": 0, "kept": 2},
 "containers": {"orphans": 0, "removed": [], "in_grace": 0, "missing_rows": 0},
 "databases": {"schemas": 14, "orphans": 3, "names": {"jpltob7s3k": [...]}},
 "deploy_records": {"benches": 0, "deleted": 0, "cap": 50},
 "verified": {"rechecked": 0, "reissued": 0, "reaped": 0, "still_present": []}}
```

`BenchPress Settings` gains `deploy_log_cap` (50). `SYSTEM_SCHEMAS` becomes
`NOT_SITE_DATABASES` and excludes `backups`, a directory MariaDB lists because backups are
written into its data directory — a report that named it could never reach zero.

### How to verify

```bash
git grep -n 'benchpress\.ingress\.\(reconcile\|enqueue_route_reconcile\)'   # empty
docker compose exec backend bench --site frontend migrate
docker compose restart backend queue-long queue-short scheduler websocket
docker compose exec queue-long bench --site frontend execute benchpress.reconcile.run
```

Run it from `queue-long`, not `backend`: `backend` has the Docker socket but not the route
mount, so the routes step reports `null` counts there rather than converging.

### Verified against the live host

- **Scheduler row.** One `Scheduled Job Type` names `benchpress.reconcile.enqueue_run`,
  none name the old path, and every `benchpress.*` method resolves under `frappe.get_attr`.
- **The reap, by the cron itself.** A hand-made container labelled
  `benchpress.managed=true` was planted at 06:22 and removed at the 06:40 tick — spared by
  the 06:32 and 06:36 ticks, because its 15-minute grace expired at 06:37. No hand-run.
- **The negative control.** An unlabelled container planted at the same moment was still
  running afterwards. `list_benches` filters on the label at the daemon, so it never
  entered the candidate set.
- **A deploy in flight.** 330 polls of the full `run()` across two deploys of a throwaway
  bench. The new container was first seen at t=10.7 s and classified `in_grace`, never an
  orphan; `removed_ever` stayed empty and both deploys reached `Running`. Polling inline
  rather than through RQ deliberately removes the single-worker serialisation that
  normally keeps this pass off a deploy.
- **A removal that did not take.** With `remove_container` patched to raise, the pass
  reported `orphans: 1, removed: [id]` and then
  `verified: {rechecked: 1, reissued: 1, reaped: 0, still_present: [id]}`, wrote one Error
  Log row, and the container was still there. Unpatched, the next pass reported
  `reaped: 1` and the container was gone.
- **Orphan databases.** 14 schemas against the claimed set, 3 unclaimed, named per server.
  No `DROP` was issued — the test asserts that, and it is the one that keeps a future
  contributor from "finishing" the feature.
- **`kept: 2`** on every pass; `dynamic.yml` (Aug 19) and `wildcard-anchor.yml` (Aug 24)
  unmodified, `traefik` never restarted, `IP Allocation` back to its baseline of 16.

### Tests

`test_reconcile` 43 (19 integration + 24 unit), `test_lifecycle` 38, `test_docker_manager`
71, `test_vpn_adapter` 17 — all exit 0. Also run because the diff touches them:
`test_ingress` 67, `test_mariadb_manager` 37, `test_scheduler_entries` 5,
`test_deploy_manager` 58, `test_run_history` 11, `test_api` 58, `test_placement` 34,
`test_docker_events` 45. `uvx pre-commit@4.3.0 run --all-files` clean.

PR #261, opened 2026-08-28
Implemented phase six

Venkateshvenki404224:
Phase 1 of the leadtype-docs spec: one page through the whole pipe, before forty exist.

One MDX source now compiles for three readers — a rendered page, a coding agent
reading `AGENTS.md` from a clone, and an HTTP agent fetching `llms.txt`.

## What this adds

- `leadtype@0.4.3` and five npm scripts, so the score floor and the base URL are
  visible to a maintainer instead of buried in a workflow step.
- `docs/docs.config.mjs` — product, the three-track navigation and the `llms.txt`
  body. The operator and reference tracks carry empty `children` until their first
  pages land, because a navigation entry naming a missing page is a `config-link`
  error.
- `docs/index.mdx` and `docs/user/quick-tour.mdx`, with one screenshot captured
  from live staging.
- The landing page moved to `site/`. The design brief, the handoff and the
  screenshot checklist moved to `docs/internal/`, and the checklist is rewritten
  for the new image layout.
- A CI `Docs` job: both generate modes, lint, score against a floor of 100, and a
  drift check that fails when a generated file is stale or hand-edited.

## Four things the pipeline forced

The spec predicted one trap. Building it found four more.

1. **`leadtype lint docs` reads every `.md` under `docs/`, not only `.mdx`.** The
   16 legacy guides fail its schema — 23 errors. They are skipped through
   `lint.ignore` in the config until the page replacing each one lands.
2. **`leadtype score` reads the same tree and has no ignore.** Six unlabeled code
   fences in the legacy guides cost the whole GEO signal, capping the score at 90.
   They are labeled.
3. **Any `--include`/`--exclude` on `generate` disables the config navigation.**
   Verified against both the single-collection and the collections config shape.
   So the tracks and the legacy-file exclusion cannot both be had. Navigation won:
   until phase 6 deletes the last legacy guide, `AGENTS.md` lists them under
   "Other" with links to mirrors that generation never wrote. See below.
4. **Three generated artifacts stamp a build clock into their body.**
   `agent-readability.json`, `search-content.json` and `search-index.json` can
   never be byte-identical between two runs, so they are no longer tracked. Every
   generated file that is reproducible stays committed. Page markdown, `llms.txt`
   and `sitemap.xml` are stamped from git history instead, which is why the CI
   checkout is unshallowed.

## Known, and self-resolving

`docs-bundle/AGENTS.md` lists the 17 legacy guides under "Other" and links to
`./docs/<name>.md` files that do not exist in the bundle. No flag or config
setting suppresses that listing — it is scanned independently of every filter.
The only real fix is moving the legacy `.md` out of `docs/`, which the phase spec
rules out ("they stay until the page that replaces each one exists"). Each entry
disappears as its replacement page lands, and the last of them goes in phase 6.

## How to verify

```bash
npm ci --ignore-scripts
npm run docs:build && npm run docs:lint && npm run docs:score; echo "exit=$?"
git status --porcelain -- docs-site docs-bundle    # empty: no drift
```

The score gate was proved to have teeth, not merely observed passing: blanking the
`description` in `docs/user/quick-tour.mdx` drops it to 85/100 and exits 1, naming
description coverage. It was then restored.


Venkateshvenki404224:
## Phase 2 — the deploy path, and a demo set worth photographing

Four new user pages, a tracked seed script, and 12 screenshots plus one animation
captured from live staging. `docs/creating-labs.md` is replaced and its five
inbound links repointed in the same commit.

| Page | What it covers |
|---|---|
| `user/deploy-from-template` | The catalog, the two buttons on a card, the eleven pipeline steps as a table, and an animation of a real run |
| `user/create-a-lab` | The New lab form, and what is baked into the image against what a deploy makes |
| `user/lab-detail` | Every field, and why `status` and `container_health` disagree |
| `user/lifecycle` | The five actions, what each one keeps, and who may run them |

The pages are written against deploys that actually ran. `crm-demo` reached
Running in **21 seconds** — the per-step times in `deploy-from-template` are that
run's, and step 7 took 15 of the 21.

### How to verify

```bash
npm run docs:build && npm run docs:lint && npm run docs:score; echo "exit=$?"
```

- `docs:lint` → `All 7 files pass.` · `docs:score` → `100/100` · exit 0
- Rebuild leaves `git status --porcelain -- docs-site docs-bundle` empty
- 6 pages, shortest alt text 310 characters, largest PNG 156 KB, GIF 158 KB

Seed script, run against the live site:

```bash
python3 scripts/seed_docs_demo.py                       # twice: 0 change(s), exit 0
python3 scripts/seed_docs_demo.py --destroy --lab tier-knobs   # refuses, exit 1
```

`--destroy` refuses a mixed batch wholesale, so `--lab crm-demo --lab pids-default`
removes nothing.

### The one thing worth reviewing closely

A demo lab cannot deploy from another lab's image: `_prepare_lab_image` demands
`benchpress/<lab_id>:lab` exactly, and throws rather than building. So the seed
adds a **second tag** to an image the host already has — a pointer, no layers, no
build. `--destroy` untags only after checking the alias still resolves to the same
image ID as its source, so it cannot delete a real image.

### Noted, not fixed

`https://staging.benchpress.cloud/socket.io/` returns **502** while the websocket
container is listening on 9000. `DeployDialog.vue` and `LabDetail.vue` are driven
entirely by the `bench_deploy_log` socket event, so the deploy dialog never
updates and sits on "Waiting for the worker to pick the deploy up…" for the whole
run. That is edge routing, outside this phase. The animation was built from a
reload sequence of the Deploy log tab instead, so every frame is a real render of
the real run.

Venkateshvenki404224:
## Phase 3 — getting into a running bench, and the credits window

Seven new user pages, taking the user track to twelve plus the docs index (13 `.mdx`).

| Page | Covers |
|---|---|
| `user/vpn-devices` | registering a machine, the QR and `.conf`, device status, the connection test |
| `user/open-your-site` | the Sites card, the three states of its **Open** button, signing in as `Administrator` |
| `user/connect-ssh-vpn` | the connection card, both addresses, the SSH command, the three passwords |
| `user/code-server` | the browser IDE, its password dialog, sharing a session, restarting it |
| `user/logs-and-monitoring` | the deploy stepper, the raw log and its step markers, the build log, the resource bars |
| `user/leases-and-credits` | the countdown, the renew dialog, the meter, the ledger, the gateway handoff |
| `user/troubleshooting` | one symptom index across the whole user track |

### The credits window

`enable_credits` was set to 1, every lease and credit frame was captured in a single pass, and it was set back to 0 inside the same iteration. It reads `0` now, verified in a fresh process.

Eight frames came out of that window, including `08-countdown.gif` (291 KB, 12.7 s, cropped to the Container card). Razorpay has no keys, so the purchase is documented up to the handoff — `07-checkout-handoff.png` is the real refusal. The ledger, meter and statement were photographed after an admin credit adjustment.

### Secrets

Every frame carrying a secret belongs to a record that no longer exists, re-verified this iteration:

| Frame | Secret | Record | State |
|---|---|---|---|
| `connect-ssh-vpn/02-revealed.png` | SSH, admin and code-server passwords | bench `05c166d4…` | gone |
| `code-server/01-password-dialog.png` | code-server password | same bench | gone |
| `vpn-devices/03-registered.png` | WireGuard QR (a private key), `172.27.0.20` | peer "Sam — MacBook Pro" | gone |

### Replaced files

`connecting-to-benches.md`, `device-management.md` and `logs-and-monitoring.md` are deleted and every inbound link is repointed. The six frames under `docs/screenshots/lease/` were re-captured at 1280x800 as a non-admin user, one for one, so that folder is removed rather than moved.

### How to verify

```bash
npm run docs:build && npm run docs:lint && npm run docs:score; echo "exit=$?"
find docs -name '*.mdx' | wc -l          # 13
sed -n '/## User Track/,/^## Agent/p' docs-site/llms.txt   # 13 entries, each with a description
find docs/images -type f \( \( -name '*.png' -size +1500k \) -o \( -name '*.gif' -size +3M \) \)   # empty
```

Regenerating twice produces no diff, so nothing under `docs-site/` or `docs-bundle/` is hand-edited.

Venkateshvenki404224:
## Phase 4 — the operator track

Sixteen pages for whoever runs the box, plus twelve screenshots captured from live staging.

**Pages.** `index`, `prerequisites`, `install`, `wireguard-setup`, `settings-reference`, `database-server`, `backup-and-restore`, `golden-images`, `image-cache`, `users-and-roles`, `upgrading`, `production-safety`, `diagnostics`, and — marked **optional, off by default** at the top — `credits-and-billing`, `admission-and-limits` and `hosted-signup`.

**Every default was read off the live site**, not off a JSON file:

```python
import frappe
doc = frappe.get_cached_doc("BenchPress Settings")   # and "Credit Settings"
for f in frappe.get_meta("BenchPress Settings").fields:
    print(f.fieldname, doc.get(f.fieldname), f.default)
```

That read caught five fields whose stored value is `0` against a non-zero shipped default, where `0` means "use the built-in": `bench_event_settle_seconds`, `orphan_grace_minutes`, `deploy_log_cap`, `lease_max_attempts` and `lease_reclaim_seconds`. The settings reference carries a **Falls back to** column for exactly those.

**Four things the live site corrected in the drafts:**

- `BenchPress Settings` has **26** non-layout fields, not 27. The app's dialog carries 10 of them.
- A `BenchPress Admin` **cannot** open `/app/user` — that role has no `User` doctype access, so granting a role is a `System Manager` action. Found by trying it.
- The screen is **Build history**, not "Build logs", and a row tagged `golden` is a golden bake, not an image build.
- `custom_config` on the Database Server is **not** what the running server took. Live: `innodb_buffer_pool_size` 134217728 and `collation_server` `utf8mb4_general_ci`, against the block's 512 MB and `utf8mb4_unicode_ci`.

**`diagnostics.mdx` is new to the spec's page list** and was the hole ten other pages already linked to. It documents all eleven checks against a live run, including this host's two real failures — `kernel_ceilings` (`kernel.pty.max` 4096, below 9024) and `clock_skew` (19,800s, one timezone) — and a section on the four things diagnostics does **not** cover. That section is load-bearing: `queue-stops` is crash-looping here, so leases are claimed and never stopped while every diagnostic row stays green.

**Nine old `docs/*.md` deleted** — `getting-started`, `wireguard-setup`, `upgrading`, `hosted-signup`, `database-backup-restore`, `SIZING`, `production-safety`, `PRODUCTION_CHECKLIST`, `lab-access-acceptance`. Every link to them is repointed, in `README.md`, `CHANGELOG.md`, `setup.sh`, `upgrade.sh` and `mariadb_manager.py`. `docs/*.md` at the docs root is now `integration-notices.md` alone.

### How to verify

```bash
npm run docs:build && npm run docs:lint && npm run docs:score; echo "exit=$?"   # 0
find docs -name '*.mdx' | wc -l                                                 # 29
git grep 'docs/getting-started.md'                                              # no hits
```

Lint passes 30 files, score is 100/100, and regenerating after the artifact commit produces no diff. `enable_credits` reads `0` — this phase documents credits without switching them on.

Venkateshvenki404224:
## Phase 5 — the reference track

Eleven pages, so the docs tree is now the full 40: `index`, `architecture`,
`data-model`, `api`, `deploy-pipeline`, `lifecycle-and-events`, `networking`,
`realtime`, `configuration`, `cli-and-scripts`, `glossary`.

Tables, not pictures — this track adds no images.

### Accounting, because a reference page is only worth its coverage

- **All 50 `@frappe.whitelist()` functions are listed.** None omitted. 35 in
  `api.py`, 11 document methods, and 4 module functions in `waitlist.py` and
  `signup.py`.
- **All 20 DocTypes are listed**, checked against the live site rather than
  against the directory listing.
- Devices are named as `VPN Peer` records in `vpn_management`, with a section
  saying there is no Device DocType, so that search ends somewhere.

### llms.sections is now written, not defaulted

`llms.txt` carries all three tracks and a description for every page, plus a
task-to-page routing block. "How do I connect over SSH" resolves to
`/docs/user/connect-ssh-vpn` from the routing block and from the User Track
list.

`docs/index.mdx` still claimed the operator and reference tracks were
unpublished. Corrected.

### Verify

```bash
npm run docs:build && npm run docs:lint && npm run docs:score; echo "exit=$?"
# exit=0, "All 41 files pass.", 100/100

grep -rn "@frappe.whitelist" --include=*.py benchpress/ | wc -l   # 50, all listed
grep -c "^## " docs-site/llms.txt                                 # 6 sections, 3 tracks
wc -w < docs-site/llms-full.txt                                   # 47,940
```

A third `docs:build` produces no diff, so the committed artifacts are
reproducible.

Venkateshvenki404224:
## Phase 6 — the README, the runbook, and putting staging back

The last phase of the leadtype-docs spec. No new documentation pages: 40 is the
final count. What changed is the front page, the agent entry point, and the state
of the machine the screenshots came from.

### README: 1,044 lines to 355

Nine sections survive, and they are the ones a front page needs — what BenchPress
is, the problem, the solution, Architecture, Features, four screenshots, the link
table into all three tracks, contributing and licence.

Twelve sections move out, because `docs/` already carries each in full and a
second copy disagrees with the first the moment one of them moves: Installation,
Usage Workflow, Development Setup, Project Structure, API Reference, Data Model,
Networking, Real-Time Communication, Configuration Reference, Supported Frappe
Apps, VPN Device Management and Prerequisites.

Architecture and Features stay inline on purpose. Those two are the pitch.

**The architecture diagram was wrong and is now right.** The old one named no
Traefik and put every bench on `172.30.0.0/24`. It now matches
`docs/reference/networking.mdx`: a bench bridge under `10.20.x`, a public address
through Traefik, a tunnel address through WireGuard, and one shared MariaDB. That
disagreement is exactly what this spec exists to remove.

The four-image strip is the arc a new user walks — the Overview dashboard, a
deploy mid-pipeline, a running bench, and code-server. Every one carries a full
sentence of alt text and a prose restatement, so the strip can be deleted without
losing a fact.

### `AGENTS.md` at the repo root

A pointer at `docs-bundle/AGENTS.md`, which is generated from this checkout and
therefore matches the code beside it, plus a task-to-page routing table and the
eight rules an agent gets wrong here — including the one that costs a debugging
session, that `leadtype lint --src .` lints its own generated output.

### The `docs.benchpress.cloud` runbook lives in `benchpress_devops/PRODUCTION.md`

Publishing the site is deliberately a human's job: Traefik reloads its dynamic
directory on mtime, so a malformed file there is an outage of the product, not of
the docs. The recipe is committed rather than left in `specs/`, which is
gitignored and deleted on completion.

It records the trap that costs the first attempt. A router file dropped into
`config/traefik/generated/dynamic/` survives at most five minutes — the `*/5`
route reconcile treats every `*.yml` there as a bench route and withdraws any
whose stem is not a live instance id, and `PROTECTED_ROUTE_FILES` holds only
`dynamic.yml` and `wildcard-anchor.yml`. The router belongs in the control
plane's own `dynamic.yml`, rendered from the tracked template.

### Thirteen images deleted

`docs/images/*.png` were the old README's screenshots. Every one is superseded by
a frame under `docs/images/<track>/<page>/` captured at 1280x800 from live
staging, and none was referenced anywhere else.

### Teardown — the negative control on the whole spec

The claim this phase makes is that the loop removed exactly what it created and
nothing else. It is checkable, and it was checked.

| Check | Result |
|---|---|
| Demo labs `crm-demo`, `erpnext-training`, `helpdesk-sandbox` | gone |
| Their bench instances | gone (2 deleted) |
| Demo users `demo@` and `intern@benchpress.cloud` | gone |
| Demo WireGuard peers | gone — 18 peers before, 16 after, and the two removed are the two demo bench peers |
| Demo route files under `config/traefik/generated/dynamic/` | gone — 11 files before, 9 after, both protected files intact |
| Labs that predate this loop | all 7 present: `client-frappe-16`, `crm`, `erpnext`, `password-manager`, `pids-default`, `tier-knobs`, `vpn-management-16` |
| `enable_credits` | `0` |

No secret in any committed image belongs to a record that still exists. Three
frames carry one — `connect-ssh-vpn/02-revealed.png` and
`code-server/01-password-dialog.png` show bench `05c166d4…`, and
`vpn-devices/03-registered.png` shows the WireGuard QR for peer "Sam — MacBook
Pro" at `172.27.0.20`. All three records were queried on the live site and none
exists. The three password strings and the four `.env` secrets were also grepped
across `docs/`, `docs-site/`, `docs-bundle/`, `README.md` and `AGENTS.md`, and
none appears in text.

### How to verify this

```bash
npm run docs:build && npm run docs:lint && npm run docs:score; echo "exit=$?"
git diff --exit-code -- docs-site docs-bundle    # no generated file was hand-edited
wc -l README.md                                  # 355
grep -n '^## ' README.md                         # nine sections, no moved one among them
```

Every relative link in `README.md` (55 paths) and `AGENTS.md` (8 paths) resolves
on disk.

PR #359, opened 2026-09-05
docs(commands): name bp_test_site in the two run-tests examples

### LINKED ISSUE #357: fix(docs): CLAUDE.md and the operator docs tell you to run tests on `frontend`
Two human-facing surfaces still carry the command #356 removed from the unattended runner:

```
# CLAUDE.md:211

# docs/reference/cli-and-scripts.mdx:60
```

`CLAUDE.md` is the file every agent session loads, and its own `.agents/prompts/00_app_context.md`
says the opposite:

> Do not name `frontend` in any `run-tests` command; `bp_test_site` locally and `test_site` in CI
> are the sites that carry `allow_tests`.

`docs/reference/cli-and-scripts.mdx` is published, so it tells a self-hoster to run the command
against their own live site.

The command cannot do useful work either way — `frontend` carries no `allow_tests`, so the runner
refuses it. It refuses *after* disabling that site's scheduler from a value it holds in memory and
restores only on a clean exit. A site with the scheduler off composes mail and never sends it, with
nothing in any log to say so.

## Wanted

- Both lines name `bp_test_site`.
- `docs-site/` and `docs-bundle/` are regenerated in the same PR, not hand-edited.
- `benchpress/tests/test_runner_targets.py` (added in #356) grows `CLAUDE.md` and
  `docs/reference/cli-and-scripts.mdx` in its scanned set, so this cannot come back. Its
  `BENCH_ON_FRONTEND` pattern matches a bench invocation and not prose, so the rule can still be
  stated in words.

The `--site frontend migrate` lines in the same two files are correct and should stay: migrating
your own site is what a person does.

Split out of #356, which fixed the unattended-runner surfaces only.

Venkateshvenki404224:
Closes #357. Stacked on #358, which added the guard test this change grows.
Review that one first, and set the base of this one to `version-16` after it merges.

`CLAUDE.md` is loaded by every agent session, and
`docs/reference/cli-and-scripts.mdx` is published. Both printed a `run-tests`
command against `frontend`, the deployment's own site.

`frontend` carries no `allow_tests`, so the runner refuses it — after it
disables that site's scheduler, from a value it holds in memory and restores
only on a clean exit. A site with the scheduler off composes mail and never
sends it, and no log says so.

## What changed

- Both lines name `bp_test_site`, and each file states the rule in words beside
  the block.
- `benchpress/tests/test_runner_targets.py` grows a second file set.
  `RUNNER_FILES` may name no site but `bp_test_site` in any bench command.
  `DOC_FILES` may name `frontend` in every command except `run-tests` — a
  person migrates, clears and inspects their own site.
- The `migrate` and `console` lines beside the changed one stay on `frontend`,
  as the issue asks.
- `.agents/prompts/00_app_context.md` loses the known-defect bullet for this.
- `docs-site/` and `docs-bundle/` are regenerated, in their own commit and after
  the source commit, so the `lastModified` stamps are the ones CI reproduces.
  Three unrelated pages carried a stamp older than the commit that last touched
  their source, so the drift check was already going to fail on the next docs
  job.

## Verification

- `python3 -m unittest benchpress.tests.test_runner_targets` — 5 tests, and the
  new one was written first. It failed naming both lines, and passes now.
- `uvx pre-commit@4.3.0 run --files ...` — clean.
- `npm run docs:lint` (41 files), `docs:score` (100/100), `docs:links` (90
  generated files). A second `rm -rf docs-site docs-bundle && npm run docs:build`
  leaves the tree clean, so the drift check has nothing to report.

### DIFF
--- a/.agents/prompts/00_app_context.md
+++ b/.agents/prompts/00_app_context.md
@@ -68,8 +68,6 @@ else runs, with somebody else's benches on it?**
 - The SPA holds **zero** `AbortController` uses, so no in-flight request is ever cancelled.
 - `CLAUDE.md:232` still says to branch from `develop`; `CONTRIBUTING.md:70-72` records that
   `develop` was abandoned at the 0.1.0 release. `version-16` is the only trunk.
-- `CLAUDE.md:208` names `--site frontend` in a `run-tests` command — that site has no
-  `allow_tests` and an interrupted run leaves its scheduler off, which silently stops all mail.
 - `CLAUDE.md:59` counts eight public pages; `benchpress/www/vs/` has held three comparison pages
   since `e6e8d32`, so the count and the routing table under it are both short.
 
--- a/.claude/skills/issue-to-phases/SKILL.md
+++ b/.claude/skills/issue-to-phases/SKILL.md
@@ -207,7 +207,7 @@ completed.
   Frappe resolves scheduled and enqueued jobs from strings at run time, so moving
   a function they name breaks them while every import still works. Rewrite every
   string in the same commit, run
-  `docker compose exec backend bench --site frontend migrate`, and restart the
+  `docker compose exec backend bench --site bp_test_site migrate`, and restart the
   workers **after** — the other order leaves a `Scheduled Job Type` row pointing
   at nothing, and that fails in complete silence. Job ids never change during a
   move, and a re-export shim is not a fallback: see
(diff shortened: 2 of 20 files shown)


Venkateshvenki404224:
Closed automatically when #358 merged and this PR's base branch was deleted. The work continues in #360, rebased onto `version-16`.

Issue #361, opened 2026-09-05
fix(tests): a whole-suite run fails on a lock that scoped runs never hit


Venkateshvenki404224:
A whole-suite run reports 27 errors that a scoped re-run of the same modules does not.

```
bench --site bp_test_site run-tests --app benchpress     # version-16 @ a25e0c2
Ran 1105 tests in 440.839s
FAILED (failures=1, errors=33, skipped=12)

bench --site bp_test_site run-tests --module benchpress.tests.test_ingress   # 36 OK
bench --site bp_test_site run-tests --module benchpress.tests.test_lease     # 68 + 46 OK
```

Every one of the 27 is the same statement:

```
SELECT name FROM `tabBench Site` WHERE name='c646ce71b4f2c27cbd4b9bc33e44d05d.localhost' … FOR UPDATE NOWAIT
(1205, 'Lock wait timeout exceeded; try restarting transaction')
```

A `FOR UPDATE NOWAIT` row lock is held across tests that run in one process, so whichever module reaches it second fails on a lock the first never released. Which modules lose depends on run order, which is why a scoped run is green and the suite is not. `references/ralph-loop.md:178-183` already describes the instability; this issue is to fix it rather than work around it.

## The reason it matters more than a flaky test

The suite exits **0** while printing `FAILED`. Combined with the inverse trap `ralph-loop.md` documents — a run whose last line reads `OK` while tests failed — neither the exit code nor the final line can be trusted alone. An unattended loop reading either signal will call a red suite green.

## Wanted

- The lock is released before the test that took it ends, or the tests that contend for that row do not share a transaction.
- A whole-suite run and a scoped run of the same modules agree.
- The runner's exit code matches its result, so CI and any unattended loop can read one signal.

Found reviewing #353. Not caused by #354.

Venkateshvenki404224:
**Correction to the diagnosis above: check for a concurrent run before treating this as a test-design fault.**

While verifying this issue I found a second process running the same suite against the same site:

```
ubuntu  python -m frappe.utils.bench_helper frappe --site bp_test_site run-tests --app benchpress
        started 14:51, still running
```

It belongs to a different session. Two suites sharing one site explains the symptom directly: one run holds the `FOR UPDATE NOWAIT` lock on `tabBench Site` while the other reaches the same row, and the loser reports a lock timeout. It also explains why a scoped re-run is green — by then the other run has moved on.

Evidence it is concurrency rather than run order:

[...]

PR #363, opened 2026-09-05
fix(tests): test isolation — lock collision and exit-code ambiguity

### LINKED ISSUE #362: fix(tests): test_deploy_manager cannot run twice on the same site
`test_deploy_manager` fails in `setUpClass`, so the whole class errors out before a single test runs:

```
IntegrityError(1062, "Duplicate entry 'test-db-steps' for key 'container_name'")
```

and the same for `test-db-notify` on `Database Server`. The count moves between runs — 26 tests/1 error, then 21 tests/2 errors — so it is order-dependent rather than a straight assertion failure.

## Cause

`setUpClass` inserts `Database Server` rows with fixed `container_name` values (`test-db-steps`, `test-db-notify`). `container_name` is unique. When a previous run left its row behind — because it was interrupted, or because the class errored before its own teardown — the next run cannot insert and every test in the class errors.

Confirmed on `bp_test_site` while reviewing #353: a row `benchpress-mariadb`, status `Active`, owner `lease-owner@example.com`, created 2026-09-02, was still present three days later. It was the only `Database Server` row on the site, referenced by no `Bench Instance`. I deleted it; `Database Server` is now empty.

That row is a symptom rather than the cause. Any interrupted run reintroduces one.

## Wanted

- The fixtures tear themselves down, so an interrupted run cannot poison the next one — `addCleanup` at insert time rather than a `tearDownClass` that a `setUpClass` failure skips.
- Or the rows carry a per-run unique `container_name`, so two runs cannot collide on the unique key.
- `test_deploy_manager` passes twice in a row from a site that already has its rows.

Found reviewing #353. Not caused by #354.

Venkateshvenki404224:
Closes #362.

## What was wrong

Three `setUpClass` methods in `test_deploy_manager.py` inserted `Database Server` rows with fixed `container_name` values (`test-db-server`, `test-db-steps`, `test-db-notify`). `container_name` is unique. Two failure modes produced the `IntegrityError (1062 Duplicate entry)` that errored the whole class before a single test ran:

- **Interrupted run**: `setUpClass` errors skip `tearDownClass`, leaving the row behind. The next run hits the duplicate.
- **Concurrent run**: a second suite running against the same site inserts while the first is still live.

## Fix

Two changes applied together to all three classes:

**1. Per-run unique `container_name`** (uuid4 suffix):
```python
container_name = f"test-db-server-{uuid.uuid4().hex[:8]}"
```
Two runs can never collide on the unique key, whether the competing row is live or stale.

**2. `addClassCleanup` at insert time**:
```python
cls.addClassCleanup(
    lambda n=cls.db_server_name: (
        frappe.delete_doc("Database Server", n, force=True, ignore_permissions=True)
        if frappe.db.exists("Database Server", n)
        else None
    )
)
```
Teardown is registered the moment the row exists, so a `setUpClass` failure cannot skip it.

The redundancy is intentional: unique names cover concurrent runs; `addClassCleanup` covers interrupted ones.

## Verification

- `uvx pre-commit@4.3.0 run --files benchpress/tests/test_deploy_manager.py` — all checks pass
- `import uuid` added to imports (stdlib, no new dependency)

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -133,7 +133,11 @@ Fifteen facts that are easy to break:
   nothing behind a page in Desk any more, so a section that should not appear is
   absent from the template rather than switched off.
 - The contact form's routing is constants too, in `benchpress/contact.py`:
-  `TOPICS` (first row is the default), `RESPONSE_TIMES` and `ACKNOWLEDGE_SENDER`.
+  `TOPICS` (first row is the default) and `ACKNOWLEDGE_SENDER`. One `TOPICS` row
+  carries the label, the address it routes to and the window the acknowledgement
+  promises, and the page's response-time panel renders those same rows. The window
+  had its own constant once, keyed by subjects that matched no label, so every
+  sender was told the same day.
   Only the forwarding address is per-deployment — the `benchpress_contact_email`
   site-config key, falling back to `CONTACT_EMAIL`.
 - The seven `Email Template` rows are the one thing `benchpress.public_site.seed`
--- a/benchpress/contact.py
+++ b/benchpress/contact.py
@@ -15,7 +15,10 @@ from benchpress.throttle import public_form
 
 DOCTYPE = "Contact Message"
 MESSAGES_PER_HOUR = 3
-SUCCESS_BODY = "Thanks — it is in front of a person, not a queue. You will hear back within one business day."
+# The mail names the window for the topic; the page says nothing narrower than every topic keeps.
+SUCCESS_BODY = (
+	"Thanks — it is in front of a person, not a queue. Most messages get an answer within one business day."
+)
 MAIL_ERROR_TITLE = "BenchPress contact mail failed"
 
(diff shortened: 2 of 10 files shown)


Venkateshvenki404224:
This commit also closes #361.

**Lock collision fix**: The `_fresh_bench` cleanup order was LIFO-wrong — `frappe.db.commit` ran FIRST, so `_delete_bench_sites` and the bench instance delete were never committed. `IntegrationTestCase` rolled them back, leaving the `Bench Site` row committed by `_run_deploy` alive. The deterministic bench name (`c646ce71b4f2c27cbd4b9bc33e44d05d` = `get_instance_id("Administrator", "test-lab-steps")`) caused a `DuplicateEntryError` → `_refuse_taken` → `FOR UPDATE NOWAIT` error 1205 in any module sharing the process that touched the same row.

Fixed by reordering: deletes first, commit last.

**Exit-code ambiguity fix**: `scripts/prompt.md` now explicitly states that agents must read `$?` not the last line of output, since `run-tests` prints two summaries and the last always reads `OK`.

